PR #10, opened 2019-12-11
'In progress -> Still contains bugs'

tchaton:


nicolas-chaulet:
Still needs some more work. I would like Unet to inherit the base model. And revisit the optimizer aspect as well.

nicolas-chaulet:
@tchaton I have changed a few things and removed non used bits. If you are good with that, feel free to merge.

nicolas-chaulet:
@tchaton something may have changed in KPConv, I have out of memory errors on the GPU now,. pointnet and RSConv work just fine

PR #15, opened 2019-12-13
Implement Residual block ABC, and concrete dilated residual block from randla-net

tristanheywood:
- Implement BaseResnetBlock in core modules, to act as the baseclass for the residual blocks from randla-net and KP conv
- move the random sampling from randlaconv to an ABC called BaseKNNConvolution
- Implement the dilated residual block from randla-net
- Implement RandLANetRes using this block - only 2 layers instead of the 4 layers from the paper. 

RandLANetRes achieves 89% training accuracy after half an epoch. 

I'm not sure about my method for dealing with the index when the convolution downsamples the pointcloud. At the moment I return an index if the pointcloud is down samples, otherwise None. It would be cleaner to always return an index (but less efficient).


models/core_modules.py
```diff
@@ -62,6 +63,89 @@ def forward(self, data):
...
+
+    def forward(self, data):
+        x, pos, batch = data
+
+        if self.ratio == 1: #convolve every point
+            row, col = knn(pos, pos, self.k, batch, batch)
+            edge_index = torch.stack([col, row], dim=0)
+            x = self.conv(x, (pos, pos), edge_index)
+            return x, pos, batch, None
+        else: #downsample using self.sampling_strategy and convolve
```
nicolas-chaulet: This else is not needed.  It will make the code more readable

models/core_modules.py
```diff
@@ -62,6 +63,89 @@ def forward(self, data):
         data = (x, pos, batch)
         return data
 
+class BaseKNNConvolution(ABC, torch.nn.Module):
```
nicolas-chaulet: A simpler solution would be to make this two helper functions that only do the sampling and not the convolution as opposed to a class, what do you think? Inheritance should be used with care

models/core_modules.py
```diff
@@ -62,6 +63,89 @@ def forward(self, data):
...
+        self.activation = ReLU()
+
+    @abstractmethod
+    def convolution(self, data):
+        pass
+
+    def forward(self, data):
+
+        x, pos, batch = data #(N, indim)
+
```
nicolas-chaulet: too many newlines here, Could you make the code a little more compact please?

models/RandLANet/modules.py
```diff
@@ -14,24 +14,17 @@ class RandlaConv(MessagePassing):
...
-        x, pos, batch = data
-        idx = torch.randint(0, pos.shape[0], (math.floor(pos.shape[0]*self.ratio),))
-        row, col = knn(pos, pos[idx], self.k, batch, batch[idx])
-        edge_index = torch.stack([col, row], dim=0)
-        x = self.propagate(edge_index, x=x, pos=(pos, pos[idx]))
-
-        pos, batch = pos[idx], batch[idx]
-        data = (x, pos, batch)
-        return data 
+    def conv(self, x, pos, edge_index):
```
nicolas-chaulet: Please see my comment rgarding changing BaseKNNConvolution to functions. Then The RandlaConv would still have a forward method.

tchaton:
One small problem to solve with BaseResnetBlock. Its subConvolution needs idx being returned as data = (x, pos, batch, idx).
Normal convolution doesn't.
Some minors modifications (make to launch training on all model to be sure nothing is broken).
The API of BaseConvolution was changed and broke for other models.

tristanheywood:
Randla-net with residual blocks gets 80% train accuracy after 60 iterations - looks like it's working

tristanheywood:
I think we should just make all convolutions return the index.

models/KPConv/modules.py
```diff
@@ -145,8 +145,11 @@ def __repr__(self):
 
 class KPConv(BaseConvolution):
     def __init__(self, ratio=None, radius=None, down_conv_nn=None, num_points=16, *args, **kwargs):
-        super(KPConv, self).__init__(ratio, radius)
+        super(KPConv, self).__init__(FPSSampler(ratio), RadiusNeighbourFinder(radius), *args, **kwargs)
 
+        self.ratio = ratio
```
nicolas-chaulet: These two properties are not used are they? If not please remove.

PR #16, opened 2019-12-13
Factory model

tchaton:


models/unet_base.py
```diff
@@ -81,11 +62,53 @@ def __init__(self, opt, num_classes, modules_lib):
 
         index -= 1
         args_up, args_down = self.fetch_arguments_up_and_down(opt, index, num_convs)
+        
         self.model = UnetSkipConnectionBlock(args_up=args_up, args_down=args_down, output_nc=num_classes, input_nc=None, submodule=unet_block,
                                              outermost=True, norm_layer=None)  # add the outermost layer
-
         print(self)
 
+    def check_if_contains_factory(self, model_name, modules_lib):
```
nicolas-chaulet: Rename to try_get_factory and return the factory or None. No need for returning a tuple

PR #18, opened 2019-12-16
Factory model

tchaton:


models/KPConv/modules.py
```diff
@@ -145,8 +145,11 @@ def __repr__(self):
 
 class KPConv(BaseConvolution):
     def __init__(self, ratio=None, radius=None, down_conv_nn=None, num_points=16, *args, **kwargs):
-        super(KPConv, self).__init__(ratio, radius)
+        super(KPConv, self).__init__(FPSSampler(ratio), RadiusNeighbourFinder(radius), *args, **kwargs)
 
+        self.ratio = ratio
```
nicolas-chaulet: Is this used? If not then please remove

PR #30, opened 2019-12-17
Better logging of loss during training

nicolas-chaulet:
Starts the work to abstract away the metrics. This is needed for allowing other tasks.


nicolas-chaulet:
#30

PR #36, opened 2019-12-18
Tracker

nicolas-chaulet:
Now we can use a metric tracker to track progress. We would have one per task and in the future we can add the logic for tracking the best model etc... For now it is just a basic implementation for the segmentation task
Solves #31


metrics/confusionmatrix.py
```diff
@@ -0,0 +1,121 @@
...
+
+    """returns list of lists of integers; use it as result[ground_truth][predicted]
+     to know how many samples of class ground_truth were reported as class predicted"""
+
+    def get_confusion_matrix(self):
+        return self.confusion_matrix
+
+    """returns list of 64-bit floats"""
+
+    def get_intersection_union_per_class(self, alpha=None, modified=False):
```
tchaton: Please, remove alpha and modified elements everywhere within this class

metrics/metricstracker.py
```diff
@@ -0,0 +1,127 @@
+import numpy as np
```
tchaton: please, rename metrics_tracker.py.

PR #39, opened 2019-12-19
Dynamicinput

nicolas-chaulet:
What's new:
- Simpler model building logic
- Model takes a whole dataset as input
- Moved methods from dataset specific implementations to the BaseDataset class
- Pointnet and KPConv can take features as an input as well as just points

### DIFF
--- a/datasets/base_dataset.py
+++ b/datasets/base_dataset.py
@@ -31,6 +31,8 @@ class BaseDataset():
         self.strategies = {}
 
     def _create_dataloaders(self, train_dataset,  test_dataset, validation=None):
+        """ Creates the data loaders. Must be called in order to complete the setup of the Dataset
+        """
         self._num_classes = train_dataset.num_classes
         self._feature_dimension = self.extract_point_dimension(train_dataset)
         self._train_loader = DataLoader(train_dataset, batch_size=self.training_opt.batch_size, shuffle=self.training_opt.shuffle,
--- a/datasets/base_dataset.py
+++ b/datasets/base_dataset.py
@@ -3,7 +3,7 @@ import logging
 
 import torch
 import torch_geometric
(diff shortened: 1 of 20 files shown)


datasets/base_dataset.py
```diff
@@ -30,31 +30,37 @@ def __init__(self, dataset_opt, training_opt):
...
-    def input_dim(self):
-        pass
+    def feature_dimension(self):
+        return self._feature_dimension
+
+    @staticmethod
+    def extract_point_dimension(dataset: Dataset):
+        sample = dataset[0]
+        if sample.x is None:
+            return 0
```
tchaton: Should it be 0 ? If none, it should be sample.pos which is used I think.

models/unet_base.py
```diff
@@ -34,50 +37,48 @@ def __init__(self, opt, model_name, num_classes, modules_lib):
...
                 self._save_sampling_and_search(unet_block, index)
         else:
             index = num_convs
 
         index -= 1
         args_up, args_down = self._fetch_arguments_up_and_down(opt, index, num_convs)
-
-        self.model = UnetSkipConnectionBlock(args_up=args_up, args_down=args_down, output_nc=num_classes, input_nc=None, submodule=unet_block,
-                                             outermost=True, norm_layer=None)  # add the outermost layer
+        args_down['nb_feature'] = dataset.feature_dimension
```
tchaton: Great :)

tchaton:
Great feature to the framework. Now, we can benchmark the different models.
Just one minor choice to make for the default input size.

datasets/base_dataset.py
nicolas-chaulet: Ok, let's do that. If no feature then x,y,z is the feature vector

PR #41, opened 2019-12-20
'working RSConv'

tchaton:


models/RandLANet/modules.py
```diff
@@ -64,7 +64,10 @@ class RandlaConv(BaseConvolutionDown):
 
     def __init__(self, ratio=None, k=None, *args, **kwargs):
         super(RandlaConv, self).__init__(RandomSampler(ratio), KNNNeighbourFinder(k), *args, **kwargs)
-
+        if kwargs.get('index') == 0 and kwargs.get("nb_feature") is not None:
+            kwargs["point_pos_nn"][-1] = kwargs.get("nb_feature")
+            kwargs["attention_nn"][0] = kwargs["attention_nn"][-1] = kwargs.get("nb_feature") * 2
+            kwargs['down_conv_nn'][0] = kwargs.get("nb_feature") * 2
```
tristanheywood: I don't really like changing the input arguments like this. Maybe we could make it so you can put "-1" for a dimension in the config if you want it to be set automatically. Otherwise I think people will be confused if the structure of the nn is not what they specify in the config.

tchaton: I totally agree, but -1 isn't satisfying either. We were thinking to have INPUT_DIM * n, but doesn't make too much sense either.

tchaton: It is even more true with troublesome model like RandALNet

tchaton: @nicolas-chaulet @tristanheywood, let's decide on slack what to do :)

tchaton: Option 1: We set every entry related dynamic input with -1, and the user needs to write the associated values in the convolution builder.
We check for -1 (if any left, we trigger an error).
It makes more sense than having a 3 which is going to be removed.

nicolas-chaulet:
Yes, -1 sounds good for a placeholder

tchaton:
Ok, I will update the code and push everything.

knn_debug.py
```diff
@@ -0,0 +1,17 @@
+from torch_geometric.nn import fps, radius, knn
+import torch
```
nicolas-chaulet: Are you sure you want to commit that file and the .p?

metrics/metrics_tracker.py
```diff
@@ -1,14 +1,70 @@
...
 import wandb
+from collections import OrderedDict
 
 from metrics.confusion_matrix import ConfusionMatrix
+from tqdm import tqdm, std
 
+HAS_TENSORBOARD_INSTALLED = False
+try:
+    from torch.utils.tensorboard import SummaryWriter
+    HAS_TENSORBOARD_INSTALLED = True
```
nicolas-chaulet: Just add tensorboard to poetry

metrics/metrics_tracker.py
```diff
@@ -1,14 +1,70 @@
...
-def get_tracker(task: str, dataset, wandb_log: bool):
+
+class COLORS:
+    TRAIN_COLOR = '\033[0;92m'
+    VAL_COLOR = '\033[0;94m'
+    TEST_COLOR = '\033[0;93m'
+    BEST_COLOR = '\033[0;92m'
+
+
+class Coloredtqdm(tqdm):
```
nicolas-chaulet: Would be better in it’s own file

metrics/metrics_tracker.py
```diff
@@ -19,17 +75,26 @@ def get_tracker(task: str, dataset, wandb_log: bool):
...
-    def __init__(self,  wandb_log: bool):
+    def __init__(self,  wandb_log: bool, use_tensorboard: bool, log_dir: str = 'logs'):
         self._wandb = wandb_log
+        self._use_tensorboard = use_tensorboard
+        self._stage = None
+        self._n_iter = 0
+
+        if self._use_tensorboard and HAS_TENSORBOARD_INSTALLED:
+            dirname = Path(os.path.abspath(__file__)).parent.parent
+            self._log_dir = os.path.join(dirname, log_dir, datetime.now().strftime("%Y%m%d-%H%M%S"))
```
nicolas-chaulet: Hydra already manages custom sur per experiment. I don’t think we need a new one. Could you remove that from the argument list?

metrics/metrics_tracker.py
```diff
@@ -43,14 +108,22 @@ def get_metrics(self) -> Dict[str, float]:
     def track(self, loss, outputs, targets):
         pass
 
+    def publish_to_tensorboard(self, metrics):
+        if self._stage == "train":
```
nicolas-chaulet: I suggest we call this method as part of the publish call

metrics/metrics_tracker.py
```diff
@@ -124,4 +196,6 @@ def get_metrics(self) -> Dict[str, float]:
         metrics['{}_macc'.format(self._stage)] = _meter_value(self._macc_meter, dim=0)
         metrics['{}_miou'.format(self._stage)] = _meter_value(self._miou_meter, dim=0)
 
+        if self._use_tensorboard:
```
nicolas-chaulet: Publishing should happen independently from get_metrics. Please remove the lines below

models/losses.py
```diff
@@ -0,0 +1,104 @@
+import numpy as np
+from typing import *
+from collections import OrderedDict
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.autograd import Variable
+
+
+def filter_valid(output, target, other=None):
```
nicolas-chaulet: Not sure we need that do we?

models/RandLANet/modules.py
tristanheywood: I think we have a placeholder (-1, NaN, None, or something like _ if underscores are valid syntax) and the user can either put the placeholder or a number. If they put a number, we check that it matches the dimensions and throw an exception if not.

tristanheywood: It would be nice to have variables and arithmetic with variables in the model definitions.

knn_debug.py
tchaton: No, I forgot to remove them.

metrics/metrics_tracker.py
tchaton: Done

metrics/metrics_tracker.py
tchaton: Yes, done

metrics/metrics_tracker.py
tchaton: Done, but it was nice to have every batch logs

models/losses.py
tchaton: We might need it, as S3DIS have some points which are un-labelled. It should be used by default to prevent this to happen.

metrics/metrics_tracker.py
tchaton: Not sure what you are referring to.

nicolas-chaulet:
This!
https://hydra.cc/docs/tutorial/working_directory/

metrics/metrics_tracker.py
nicolas-chaulet: https://hydra.cc/docs/tutorial/working_directory/

models/losses.py
nicolas-chaulet: Then it should be in the dataset as a filter no?

metrics/metrics_tracker.py
tchaton: Will do that in the next PR.

PR #43, opened 2019-12-23
Rs conv placeholder

tristanheywood:
This is proposed solution for using placeholders in the config to stand in for feature dimensions. 

I put a variable called FEAT in the config file, without quotes, to stand in for the feature dimension. Hydra automatically converts this to a string. I evaluate this string using eval, and providing a local variable called 'FEAT'. This allows arithmetic to be done with variables in the config. 

So far I have only implemented this for one RSConv model, if you guys are happy with this idea we can do it for all models. 

If we use this eval method to resolve all mlp lists, it will let users do arithmetic between numbers in the config file (e.g [64, 64 + 64]).


tristanheywood:
Unit tests are failing because I changed FPModule to use the placeholder, but only changed the config for RSConv_2LD

PR #44, opened 2019-12-26
Model checkpoint

tchaton:
This PR adds a ModelCheckpoint Class which will handle all the saving logic.
Currently, it will be used during the publish function of the BaseTracker.

Checkpoint state:
        self._objects = {}
        self._objects['models'] = {}
        self._objects['stats'] = {'train': [], 'test': [], 'val': []}
        self._objects['optimizer'] = None
        self._objects['scheduler'] = None
        self._objects['args'] = None
        self._objects['kwargs'] = None
        self._filled = False

models = ['default'] + one per best metric on the test.

- [x] Create a Checkpoint Object
   - [x] load_state
   - [x] save_state

- [x] Update tracker to call ModelCheckpoint
   - [x] initialize from checkpoint 
   - [x] save tracked elements


metrics/metrics_tracker.py
```diff
@@ -24,28 +27,52 @@ def get_tracker(task: str, dataset, wandb_log: bool, use_tensorboard: bool, log_
...
-    def __init__(self,  wandb_log: bool, use_tensorboard: bool, log_dir: str = 'logs'):
+    def __init__(self, wandb_log: bool, use_tensorboard: bool, log_dir: str = 'logs', experiment_name: str = None, use_checkpoint: bool = True):
         self._wandb = wandb_log
         self._use_tensorboard = use_tensorboard
         self._stage = None
         self._n_iter = 0
+        self._use_checkpoint = use_checkpoint
+        self._default_metric_to_func = {'acc': max, 'iou': max, 'loss': min}
+
+        dirname = Path(os.path.abspath(__file__)).parent.parent
```
nicolas-chaulet: Please see https://hydra.cc/docs/tutorial/working_directory/
I think you can remove `log_dir` and put everything related to logs in the current folder which is going to be the hydra experiment folder.

nicolas-chaulet: The problem with the implementation here is what we log things at two different places the Hydra location and this dirname custom location. There should only be one for all the logs related to a training.

tchaton: Ok :)

To do
- [ ] Update output saving using hydra. 

Model checkpoint: I am doing exactly that: "for tracking best models for different metrics for example". I am using the metrics being published to track improved models on the test. 

- [ ] Make ModelCheckpoint / Tracker more indepent

tchaton: I still need to add more tests, but the training looks ok :)

tchaton:
Screenshot from 2019-12-28 12-21-39

.github/workflows/pythonapp.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         python -m pip install --upgrade pip
         curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
         $HOME/.poetry/bin/poetry install
-    - name: Lint with flake8
+    - name: Link with flake8
```
nicolas-chaulet: It is actually Lint!! :)

Issue #46, opened 2019-12-27
Improve placeholders and variables in the model config


tristanheywood:
Right now the only placeholder supported is 'FEAT', the number of input features. This placeholder needs to be resolved by each model separately. 

It would be better to resolve all placeholders as part of the core of the framework, so that model implementations don't need to worry about it. We could also allow users to do arithmetic in any of the fields (.e.g nn: [64, 64*2]). 

We could also have a 'define_placeholders' section of the config where users can create their own placeholders. E.g. 

```
MyModel:
    type: RSConv
    define_placeholders:
        L1_OUT: 64
    down_conv:
         nn: [[FEAT, 32, L1_OUT], [L1_OUT, L1_OUT*2, 64]
```

nicolas-chaulet:
Even if the idea is quite appealing I am not 100% convinced that placeholders add something or simplifies the configuration of models for the following reasons:

- For inputs, the number of features is actually completely fixed by the dataset and the selected  model (nothing the user can control). For example for PoinNet++ it is `3 + number_features` but for KPConv it is `number_features`. Size of input layers is not configurable, that's why I liked the -1 approach that then gets replaced by each model independently. 

- When you have sequences of MLP it is a bit the same, in the example above the input to the second MLP (L1_OUT) is totally defined by the output of the first MLP, they have to match anyway. So again, not something that the user has any control over. The size of the second layer being 2*L1_OUT would add some value though. But not sure it is worth maintaining a whole system of regular expressions for that. I think it's ok to write your MLPs every time entirely

Bottom line I am not convinced it will help users while certainly adding complexity when first getting into the framework. Shall put it aside until we have some more people weighing in and see where it goes?

tristanheywood:
You raise some good points, but

- The user can control which dataset they use with a given model, so they need some way of having the same model deal with different numbers of features. 'FEAT' is determined by the dataset, so for pointnet++ you'd write "nn: [3 + FEAT, ...]" and for KPConv "nn: [FEAT, ...]". I think it's good to make the config as transparent as possible. I found it very confusing when I specified one number for a MLP in the model, but the model printout showed a different number. 
- It wouldn't be implemented using regular expressions. The method I used on the RSConv branch was to call eval() on the string, with 'FEAT' as a local variable. This method allows for any arithmetic operations involving both placeholders and number literals and avoids complicated parsing logic. 

[...]

PR #47, opened 2019-12-28
Add expressions and constants to model definition (globally)

tristanheywood:
Solves #46 

This lets users put things like `inDim: FEAT` or `nn:[FEAT*2, 16 + 16]` in the model definition. Previously these expressions had to be evaluated per argument within the convolutions. Now it is done for the entire model config before calling `find_model_using_name`

### DIFF
--- a/conf/models/segmentation.yaml
+++ b/conf/models/segmentation.yaml
@@ -39,7 +39,7 @@ models:
             nn: [131, 128]
         up_conv:
             module_name: FPModule
-            up_conv_nn: [[256, 128], [160, 64], [64 + FEAT, 64]]
+            up_conv_nn: [[128 + 128, 128], [128 + 32, 64], [64 + FEAT, 64]]
             up_k: [1, 1, 1]
             skip: True
         mlp_cls:
@@ -61,7 +61,7 @@ models:
             nn: [131, 128]
         up_conv:
             module_name: FPModule
-            up_conv_nn: [[256, 128], [160, 64], [80, 64], [64 + FEAT, 64]] 
+            up_conv_nn: [[128 + 128, 128], [128 + 32, 64], [64 + 16, 64], [64 + FEAT, 64]] 
             up_k: [1, 1, 1, 1]
             skip: True
         mlp_cls:
@@ -106,7 +106,7 @@ models:
             nn: [131, 128] #[3  + 128]
         up_conv:
             module_name: FPModule
-            up_conv_nn: [[256, 128], [192, 64], [96, 32], [48, 32], [32, 64]]
+            up_conv_nn: [[128 + 128, 128], [128 + 64, 64], [64 + 32, 32], [32 + 16, 32], [32, 64]]
             up_k: [1, 3, 3, 3, 3]
             skip: True
         mlp_cls:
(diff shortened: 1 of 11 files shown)


conf/models/segmentation.yaml
```diff
@@ -80,12 +80,12 @@ models:
...
             module_name: GlobalBaseModule
             aggr: max
-            nn: [131, 128] #[3  + 128]
+            nn: [128 + FEAT, 128] 
         up_conv:
             module_name: FPModule
             ratios: [1, 0.25, 0.2]
             radius: [0.2, 0.2, 0.1]
-            up_conv_nn: [[256, 64], [128, 64], [64 + FEAT, 64]]
+            up_conv_nn: [[128 + 128, 64], [64 + 64, 64], [64, 64]]
```
tchaton: Add this to every models please

models/model_building_utils/model_definition_resolver.py
```diff
@@ -0,0 +1,48 @@
+
+from omegaconf.dictconfig import DictConfig 
+from omegaconf.listconfig import ListConfig
+
+def resolve_model(model_config, dataset, tested_task):
+
+    #placeholders to subsitute 
+    constants = {
+        'FEAT': dataset.feature_dimension, 
+        'TASK': tested_task
```
tchaton: Why TASK is used there ?

test/test_models.py
```diff
@@ -9,7 +9,12 @@
 
 from models.utils import find_model_using_name
 from models.pointnet2.nn import SegmentationModel
+from models.model_building_utils.model_definition_resolver import resolve_model
 
+#calls resolve_model, then find_model_using_name
+def _find_model_using_name(model_type, task, model_config, dataset):
```
tchaton: Swith default find_model_using_name to use resolve_model by default. The other one should be with _

train.py
```diff
@@ -85,6 +86,7 @@ def main(cfg):
     # Find and create associated model
     model_config = getattr(cfg.models, tested_model_name, None)
     model_config = OmegaConf.merge(model_config, cfg.training)
+    resolve_model(model_config, dataset, tested_task)
```
tchaton: move this into find_model_using_name

models/model_building_utils/model_definition_resolver.py
tristanheywood: The XConv from PointCNN is slightly different for classification vs segmentation, so I plan to provide the task as an argument to the convolution.

train.py
tristanheywood: I put it in train.py so it's more obvious to people looking at the code where this resolving is happening. But I can move it if you think that's cleaner.

test/test_resolver.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+import sys
+ROOT = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..')
+sys.path.append(ROOT)
+
+from models.model_building_utils.model_definition_resolver import resolve_model
+
+
+
+class MockDataset(torch.utils.data.Dataset):
```
nicolas-chaulet: Since this dataset is used across multiple tests it makes sense to take it out in its own file

train.py
nicolas-chaulet: I think I agree with Tristan on that one

tchaton: Yes, agree :)

PR #48, opened 2019-12-28
PointCNN + layer list api for model definition

tristanheywood:
This PR implements PointCNN from https://arxiv.org/pdf/1801.07791.pdf . I trained for a few iterations and saw a slight increase in accuracy, but I don't have access to a GPU at the moment so I can't be sure this implementation is good. 

- XConv from pytorch_geometric, adapted for this framework 
- Wrapper classes for up and down convolutions based on XConv
- Dilated KNN neighbour finder (find K * D neighbours and choose K of them at random) 
- layer list api for model definition - the model is specified by a list of convolutions, rather than a convolution with a list of params. I added this with PointCNN because XConv has a lot of arguments.  

At the moment there's one function for building a Unet from the current api, and one for building one from the layer list api. In the future we should combine these into one function. I've written a function which can flatten the current api into the layer list.


conf/models/segmentation.yaml
```diff
@@ -22,6 +22,109 @@ models:
             nn: []
             dropout: 0.5
 
+    pointcnn_small:
```
tchaton: It is great to have both. I think it is harder to read as it is spread out though.

models/core_sampling_and_search.py
```diff
@@ -64,3 +74,24 @@ def __init__(self, k):
...
+        self.k = k 
+        self.dilation = dilation
+        self.initialFinder = KNNNeighbourFinder(k * dilation)
+
+    def find_neighbours(self, x, y, batch_x, batch_y):
+        #find the self.k * self.dilation closest neighbours in x for each y
+        row, col = self.initialFinder.find_neighbours(x, y, batch_x, batch_y) 
+
+        #for each point in y, randomly select k of its neighbours
+        index = torch.randint(self.k * self.dilation, (len(y), self.k), device = row.device, dtype = torch.long)
```
tchaton: torch.randint can doesn't have a replace=True option, meaning it can generate several time the same value, and therefore you will end up selecting < k distinct neighbours.

I would prefer something like (might be faster too)
perm = torch.randperm(tensor.size(0))
idx = perm[:k]

I have made some research. randperm doesn't support broadcasting. So it is possible to use it, with %modulo, but it might be slower than yours. 
I guess having duplicated points is acceptable for now.

models/unet_base.py
```diff
@@ -47,6 +48,18 @@ def __init__(self, opt, model_type, dataset: BaseDataset, modules_lib):
         It is a recursive process.
         """
         super(UnetBasedModel, self).__init__(opt)
+        #detect which options format has been used to define the model 
+        if type(opt.down_conv) is ListConfig:
```
tchaton: Could use: type(opt.down_conv) is ListConfig and type(opt.up_conv) is ListConfig

tristanheywood: I did this so you can use the layer list api for the down conv, and the normal api for the up conv. It will all be cleaned up if we merge the two model building functions.

models/core_sampling_and_search.py
tristanheywood: yeah it would be nice to sample without replacement. We can look into this in the future.

PR #63, opened 2019-12-31
Fix shapenet

nicolas-chaulet:


datasets/shapenet_dataset.py
```diff
@@ -199,9 +200,10 @@ def __init__(self, dataset_opt, training_opt):
         self._data_path = os.path.join(dataset_opt.dataroot, 'ShapeNet')
         self._category = dataset_opt.category
         pre_transform = T.NormalizeScale()
+        transform = T.FixedPoints(dataset_opt.num_points)
```
tchaton: Did you check Normalize ? It should center the points too like :

```
def pc_normalize(pc):
    l = pc.shape[0]
    centroid = np.mean(pc, axis=0)
    pc = pc - centroid
    m = np.max(np.sqrt(np.sum(pc**2, axis=1)))
    pc = pc / m
    return pc
```

nicolas-chaulet: Isn't this what we have in the pre-transform just above?

PR #67, opened 2020-01-01
Multiscale pointnet

nicolas-chaulet:
Support for multi scale pointnet.
Also introduces a new base conv for multi scale convolutions.

### DIFF
--- a/conf/models/segmentation.yaml
+++ b/conf/models/segmentation.yaml
@@ -269,6 +269,7 @@ models:
             ratios: [0.2, 0.25]
             radius: [0.2, 0.4]
             down_conv_nn: [[FEAT + 3, 64, 64, 128], [128 + 3, 128, 128, 256]]
+            radius_num_points: [64, 64]
         up_conv:
             module_name: FPModule
             up_conv_nn: [[1024 + 256, 256, 256], [256 + 128, 256, 128], [128 + FEAT, 128, 128, 128]]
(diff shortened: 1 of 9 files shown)


conf/models/segmentation.yaml
```diff
@@ -279,5 +280,26 @@ models:
...
             nn: [256 + 3, 256, 512, 1024]
         mlp_cls: 
-            nn: [128, 128, 128, 128, 128]
+            nn: [128, 128, 128]
+            dropout: 0.5
+    
+    pointnet2ms:
+        type: pointnet2
+        down_conv:
+            module_name: SAModule
```
tchaton: Please, keep the same name for the Module as in the original code

models/core_modules.py
```diff
@@ -64,6 +66,50 @@ def forward(self, data):
...
+            batch_obj.idx = idx
+
+        ms_x = []
+        for scale_idx in range(self.neighbour_finder.num_scales):
+            if self._precompute_multi_scale:
+                edge_index = getattr(data, "edge_index_{}_{}".format(self._index, scale_idx), None)
+            else:
+                row, col = self.neighbour_finder(pos, pos[idx], batch, batch[idx], scale_idx)
+                edge_index = torch.stack([col, row], dim=0)
+                # batch_obj.edge_index = edge_index # THOMAS, is this used anywhere at the moment?
```
tchaton: No, I don t it is used for now

models/pointnet2/modules.py
```diff
@@ -1,14 +1,21 @@
 from torch_geometric.nn import PointConv
 from models.core_modules import *
-from models.core_sampling_and_search import RadiusNeighbourFinder, FPSSampler
+from models.core_sampling_and_search import MultiscaleRadiusNeighbourFinder, FPSSampler
 
 
-class SAModule(BaseConvolutionDown):
-    def __init__(self, ratio=None, radius=None, down_conv_nn=None, nb_feature=None, *args, **kwargs):
-        super(SAModule, self).__init__(FPSSampler(ratio), RadiusNeighbourFinder(radius), *args, **kwargs)
+class SAModule(BaseMSConvolutionDown):
```
tchaton: SAModule should be kept as before, and this one should be SAModuleMSG (I think)

tchaton:
Minor changes. Feel free to merge

conf/models/segmentation.yaml
nicolas-chaulet: It is dynamical. If you give a list of radiuses it does multiscale, otherwise it does single scale

models/pointnet2/modules.py
nicolas-chaulet: It is fully dynamical The RadiusSearch and BaseMSConvolution handle the complexity of single versus multiple scales. It all depends on the parameters given (list of radius or single radius) We could drop all the single scale code since they are redundant.

nicolas-chaulet:
@tchaton I'll leave the code like that for now, I think it's ok to have a single class that handles single and multiple scales. Let me know

PR #71, opened 2020-01-03
'simple data augmentation on s3dis'

tchaton:


datasets/s3dis_dataset.py
```diff
@@ -50,9 +50,17 @@ class S3DISDataset(BaseDataset):
     def __init__(self, dataset_opt, training_opt):
         super().__init__(dataset_opt, training_opt)
         self._data_path = os.path.join(dataset_opt.dataroot, 'S3DIS')
+        pre_transform = T.NormalizeScale()
+
+        transform = T.Compose([
+            T.FixedPoints(dataset_opt.num_points),
+            T.RandomTranslate(0.01),
+            T.RandomRotate(15, axis=0),
+            T.RandomRotate(15, axis=1),
```
nicolas-chaulet: Nice one, I will add roate around z and Jitter as well. I don't think we should Normalize through. There is already a "normalized" x,y,z in the feature list

PR #76, opened 2020-01-03
Downconv pointnet

nicolas-chaulet:
Disclaimer! This is not fully tested

This PR adds the following:
- Dense samplers and radius search
- Dense down convolution
- PN2 core down convolution using dense blocks

TODO:
- Up conv using torch_points
- Get test to pass
- Cleanup the Conv2d code (maybe put in a separate file)


models/core_sampling_and_search.py
```diff
@@ -19,14 +20,14 @@ def __init__(self, ratio=None, num_to_sample=None):
...
-    def __call__(self, pos, batch):
-        return self.sample(pos, batch)
+    def __call__(self, pos, batch=None):
+        return self.sample(pos, batch=batch)
 
     def _get_num_to_sample(self, pos) -> int:
         if hasattr(self, '_num_to_sample'):
             return self._num_to_sample
         else:
-            return math.floor(pos.shape[0]*self._ratio)
```
tristanheywood: Won't this break all models which don't use a batch dimension?

tchaton:
```
Model size = 1413965
Access tensorboard with the following command <tensorboard --logdir=/home/thomas/HELIX/research/deeppointcloud-benchmarks/outputs/2020-01-04/10-30-30/tensorboard>
EPOCH 1 / 350
  4%|██                                                  | 165/4184 [00:10<04:23, 15.28it/s, data_loading=0.002, iteration=0.060, train_acc=22.42, train_loss_seg=2.153, train_macc=13.61, train_miou=5.140]
Traceback (most recent call last):
  File "train.py", line 125, in <module>
    main()
  File "/home/thomas/.cache/pypoetry/virtualenvs/superpoint-graph-job-py3.6/lib/python3.6/site-packages/hydra/main.py", line 24, in decorated_main
    strict=strict,
  File "/home/thomas/.cache/pypoetry/virtualenvs/superpoint-graph-job-py3.6/lib/python3.6/site-packages/hydra/_internal/utils.py", line 174, in run_hydra
    overrides=args.overrides,
  File "/home/thomas/.cache/pypoetry/virtualenvs/superpoint-graph-job-py3.6/lib/python3.6/site-packages/hydra/_internal/hydra.py", line 86, in run
    job_subdir_key=None,
  File "/home/thomas/.cache/pypoetry/virtualenvs/superpoint-graph-job-py3.6/lib/python3.6/site-packages/hydra/plugins/common/utils.py", line 109, in run_job
    ret.return_value = task_function(task_cfg)
  File "train.py", line 121, in main
    run(cfg, model, dataset, device, tracker, checkpoint)
  File "train.py", line 71, in run
    train(epoch, model, train_loader, device, tracker, checkpoint)
  File "train.py", line 38, in train
    tracker.track(model.get_current_losses(), model.get_output(), model.get_labels())
  File "/home/thomas/HELIX/research/deeppointcloud-benchmarks/models/base_model.py", line 81, in get_current_losses
    errors_ret[name] = float(getattr(self, name))
RuntimeError: CUDA error: an illegal memory access was encountered
```

models/core_sampling_and_search.py
nicolas-chaulet: Yes indeed... But it is already broken in a way since pos would contain all the batches together. Maybe a better interface would be to give num_points? Or batch_size? instead of pose

models/core_modules.py
```diff
@@ -228,18 +385,17 @@ def forward(self, data):
...
-        import pdb
-        pdb.set_trace()
-        x = self.nn(torch.cat([x, pos], dim=1))
-        x = self.pool(x, batch)
+        pos_flipped = pos.transpose(1, 2).contiguous()
+        x = self.nn(torch.cat([x, pos_flipped], dim=1).unsqueeze(-1))
+        x = x.squeeze().max(-1)[0]
         batch_obj.x = x
         batch_obj.pos = pos.new_zeros((x.size(0), 3))
-        batch_obj.batch = torch.arange(x.size(0), device=batch.device)
```
nicolas-chaulet: You don't need batch here

models/core_modules.py
```diff
@@ -364,6 +364,42 @@ def forward(self, data):
...
 
+class BaseDenseConvolutionUp(BaseConvolution):
+    def __init__(self, neighbour_finder, *args, **kwargs):
+        super(BaseDenseConvolutionUp, self).__init__(None, neighbour_finder, *args, **kwargs)
+
+        self._precompute_multi_scale = kwargs.get("precompute_multi_scale", None)
+        self._index = kwargs.get("index", None)
+        self._skip = kwargs.get("skip", True)
+
+    def conv(self, x, x_skip, pos, pos_skip, batch, batch_skip):
```
nicolas-chaulet: Would it be possible to use the same interface as the one we have for the DenseDownConv? Like we don't need batch nor batch_skip here for example.

models/core_modules.py
```diff
@@ -422,6 +458,34 @@ def __init__(self, up_k, up_conv_nn, nb_feature=None, **kwargs):
...
 
+
+class DenseFPModule(BaseDenseConvolutionUp):
+    def __init__(self, up_k, up_conv_nn, nb_feature=None, **kwargs):
+        super(DenseFPModule, self).__init__(None)
+
+        self.k = up_k
+        self.nn = SharedMLP(up_conv_nn)
+
+    def conv(self, x, x_skip, pos, pos_skip, batch, batch_skip):
```
nicolas-chaulet: Same as above, please remove the batch from here

models/pointnet2_customkernel/nn.py
```diff
@@ -1,50 +1,228 @@
...
+            li_xyz, li_features = self.SA_modules[i](l_xyz[i], l_features[i])
+            l_xyz.append(li_xyz)
+            l_features.append(li_features)
+
+        for i in range(-1, -(len(self.FP_modules) + 1), -1):
+            l_features[i - 1] = self.FP_modules[i](
+                l_xyz[i - 1], l_xyz[i], l_features[i - 1], l_features[i]
+            )
+        out = l_features[0]
+        out = out.view((-1, out.shape[1])).contiguous()
```
erikwijmans: This line is likely an issue. I believe out is a (B, C, N) tensor, so this view will not do what you want. There needs to be a permute(0, 2, 1) to make it (B, N, C) before the view.  

Alternatively, you can keep it (B, C, N) and use a conv instead (if labels is a (B, N) tensor, the loss will work also!)

tchaton: Yes, you entirely right. I saw it this morning and switched back to yours. Thanks for reviewing the PR. I still need to integrate our entire code base with our BaseDenseConvolution and the segmentation.yaml interface.

PR #77, opened 2020-01-04
ABC for modules with internal losses

tristanheywood:
Modules can now indicate they have internal losses by inheriting from `BaseInternalLossModule`, and implementing the abstract method `get_internal_losses`. I think this is more intuitive for users than having an instance variable which is searched for recursively. 

The method used by `BaseModel` for collecting internal losses is still similar, except now it checks if child models inherit from `BaseInternalLossModule`. I have also added a method `get_named_internal_losses` to `BaseModel`, which returns a dict of all the internal losses in the model, which will allow users to choose the weightings between different losses when calculating the overall loss.

### DIFF
--- a/models/core_modules.py
+++ b/models/core_modules.py
@@ -1,13 +1,13 @@
 from abc import ABC, abstractmethod
 import math
 from functools import partial
+from typing import Dict, Any
 import torch
 from torch import nn
 from torch.nn import Sequential as Seq, Linear as Lin, ReLU, LeakyReLU, BatchNorm1d as BN, Dropout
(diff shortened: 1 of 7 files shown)


models/core_modules.py
```diff
@@ -268,3 +268,12 @@ def forward(self, data):
         batch_obj.batch = data.batch
         copy_from_to(data, batch_obj)
         return batch_obj
+
+
+class BaseInternalLossModule(ABC):
```
nicolas-chaulet: @tristanheywood I think this would be better placed in the base_model.py file. It is not really a module

PR #84, opened 2020-01-08
'query_and_group_shadow'

tchaton:


tchaton:
Should be merged after new torch_points update

PR #109, opened 2020-01-16
'unwrapped_unet_base in progress'

tchaton:


models/RSConv_dense/nn.py
```diff
@@ -1,17 +1,18 @@
 from typing import Any
+import numpy as np
 import torch
 import torch.nn.functional as F
 from torch_geometric.data import Data
 from models.unet_base import UnetBasedModel
```
nicolas-chaulet: Please remove if it is not used

test/test_unwrapped_unet_base.py
```diff
@@ -0,0 +1,93 @@
...
+
+from models.model_building_utils.model_definition_resolver import resolve_model
+from models.unwrapped_unet_base import UnwrappedUnetBasedModel
+from utils_folder.utils import merges_in_sub, set_format
+
+
+class SegmentationModel(UnwrappedUnetBasedModel):
+    r"""
+        RSConv Segmentation Model with / without multi-scale grouping
+        Semantic segmentation network that uses feature propogation layers
```
nicolas-chaulet: No need for this docstring

models/unwrapped_unet_base.py
```diff
@@ -0,0 +1,209 @@
...
+    def _save_sampling_and_search(self, down_conv, index):
+        self._sampling_and_search_dict[index] = [
+            getattr(down_conv, "sampler", None),
+            getattr(down_conv, "neighbour_finder", None),
+        ]
+
+    def __init__(self, opt, model_type, dataset: BaseDataset, modules_lib):
+        """Construct a Unet generator
+        Parameters:
+            opt - options for the network generation
```
nicolas-chaulet: Please review the doc here, it does not match the setup

test/test_unwrapped_unet_base.py
```diff
@@ -0,0 +1,93 @@
...
+            Whether or not to use the xyz position of a point as a feature
+    """
+
+    def __init__(self, option, model_type, dataset, modules):
+        # call the initialization method of UnwrappedUnetBasedModel
+        UnwrappedUnetBasedModel.__init__(self, option, model_type, dataset, modules)
+        pass
+
+
+class MockDataset(torch.utils.data.Dataset):
```
nicolas-chaulet: Please use the mock dataset from the mockadasets.py (unless this one is specific?)

models/unwrapped_unet_base.py
```diff
@@ -0,0 +1,209 @@
...
+            raise NotImplementedError
+        else:
+            self._init_from_compact_format(opt, model_type, dataset, modules_lib)
+
+    def get_from_kwargs(self, kwargs, name):
+        module = kwargs[name]
+        kwargs.pop(name)
+        return module
+
+    def create_down_and_up_modules(self, args_down, args_up, modules_lib):
```
nicolas-chaulet: This ad the next should be private methods

models/unwrapped_unet_base.py
```diff
@@ -0,0 +1,209 @@
...
+        return factory_module_cls
+
+    def _fetch_arguments_from_list(self, opt, index):
+        """Fetch the arguments for a single convolution from multiple lists
+        of arguments - for models specified in the compact format.
+        """
+        args = {}
+        for o, v in opt.items():
+            name = str(o)
+            if isinstance(getattr(opt, o), ListConfig) and len(getattr(opt, o)) > 0:
```
nicolas-chaulet: Let's create an is_list method in the utils that checks if something is a list or a ListConfig. I have already something like that in the core_sampling_and_search.py

models/unwrapped_unet_base.py
```diff
@@ -0,0 +1,209 @@
...
+            opt - options for the network generation
+            model_type - type of the model to be generated
+            num_class - output of the network
+            modules_lib - all modules that can be used in the UNet
+        We construct the U-Net from the innermost layer to the outermost layer.
+        It is a recursive process.
+        """
+        super(UnwrappedUnetBasedModel, self).__init__(opt)
+        # detect which options format has been used to define the model
+        if type(opt.down_conv) is ListConfig or "down_conv_nn" not in opt.down_conv:
```
nicolas-chaulet: Same comment as below. If down_conv is a list this would break

models/unwrapped_unet_base.py
nicolas-chaulet: Can you add the mandatory variables for opt? (down_conv_nn etc...)

models/base_model.py
```diff
@@ -14,6 +14,21 @@
 
 log = logging.getLogger(__name__)
 
+SPECIAL_NAMES = ["radius", "max_num_neighbors"]
+
+
+class BaseFactory:
```
nicolas-chaulet: This really belongs to one of the Unets. It is specific to that not to the base model. Please move it back

models/unet_base.py
```diff
@@ -72,6 +50,11 @@ def __init__(self, opt, model_type, dataset: BaseDataset, modules_lib):
             modules_lib - all modules that can be used in the UNet
         We construct the U-Net from the innermost layer to the outermost layer.
         It is a recursive process.
+
+        opt is expected to contains the following keys:
+        * down_conv
```
nicolas-chaulet: Nice, thanks!

models/unet_base.py
```diff
@@ -190,15 +173,15 @@ def _fetch_arguments_from_list(self, opt, index):
         args = {}
         for o, v in opt.items():
             name = str(o)
-            if isinstance(getattr(opt, o), ListConfig) and len(getattr(opt, o)) > 0:
+            if is_omegaconf_list(v) and len(getattr(opt, o)) > 0:
```
nicolas-chaulet: I was not clear. My point was that it is ok if v is a normal list as well. So the function would be is_list and return true if it is an Omega list or a normal list. Same for the dict. What do you think? 

And please merge your implementation with the existing one in core_sampling_and_search:
```
def is_list(entity):
    return isinstance(entity, list) or isinstance(entity, ListConfig)
```
We only need one of those functions

models/utils.py
```diff
@@ -2,6 +2,16 @@
...
 from datasets.base_dataset import BaseDataset
+from omegaconf.listconfig import ListConfig
+from omegaconf.dictconfig import DictConfig
+
+
+def is_omegaconf_list(opt):
+    return isinstance(opt, ListConfig)
+
+
+def is_omegaconf_dict(opt):
```
nicolas-chaulet: See my comments above.

test/test_unwrapped_unet_base.py
nicolas-chaulet: @tchaton could you please remove that for your next PR?

PR #110, opened 2020-01-16
Simpler checkpoint management

nicolas-chaulet:
This PR changes the behaviour of the checkpoints (hopefully for the better).
Now, models are saved in the hydra directory and models can be loaded from the checkpoint_dir path. Also supports saving and loading the learning rate config.

### DIFF
--- a/conf/config.yaml
+++ b/conf/config.yaml
@@ -1,8 +1,7 @@
 # Those arguments within experiment defines which model, dataset and task to be created for benchmarking
 experiment:
-    model_name: pointnet2_charlesssg
-    experiment_name: "" # Wether to use hydra naming convention for the experiment
-    log_dir: "" # Wether to use hydra automatic path generation for saving training
+    model_name: pointnet2_charlesmsg
+    checkpoint_dir: "/home/nicolas/deeppointcloud-benchmarks/outputs/2020-01-14/22-56-10"
     resume: True
     task: segmentation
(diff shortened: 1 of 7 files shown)


schedulers/lr_schedulers.py
```diff
@@ -1,4 +1,9 @@
 from torch.optim.lr_scheduler import LambdaLR
+from omegaconf.dictconfig import DictConfig
+
+
+def build_basic_params():
```
tchaton: Maybe, we could use the config.yaml to initialize this one.

nicolas-chaulet: Yes, I am not sure how to go about that one...

conf/config.yaml
```diff
@@ -1,16 +1,15 @@
 # Those arguments within experiment defines which model, dataset and task to be created for benchmarking
 experiment:
-    model_name: pointnet2_charlesssg
-    experiment_name: "" # Wether to use hydra naming convention for the experiment
-    log_dir: "" # Wether to use hydra automatic path generation for saving training
+    model_name: pointnet2_charlesmsg
+    checkpoint_dir: "/home/nicolas/deeppointcloud-benchmarks/outputs/2020-01-14/22-56-10"
```
omry: you can leave this as null in the config, indicating it's there but not having an actual value.

PR #114, opened 2020-01-17
Rs conv debug

tchaton:


models/core_modules.py
```diff
@@ -76,6 +77,13 @@ def __init__(self, sampler, neighbour_finder, *args, **kwargs):
         self.sampler = sampler
         self.neighbour_finder = neighbour_finder
 
+    @property
+    def params(self):
```
nicolas-chaulet: Good idea! Please change the name to `nb_params` and add a short description as well.  Also make sure you use it in the train.py script or cleanup the train.py to remove the print on the number of elements.

models/dense_modules.py
```diff
@@ -125,17 +127,47 @@ def conv(self, pos, pos_skip, x):
...
 
     def forward(self, data):
         x, pos = data.x, data.pos
         pos_flipped = pos.transpose(1, 2).contiguous()
+
         x = self.nn(torch.cat([x, pos_flipped], dim=1).unsqueeze(-1))
-        x = x.squeeze().max(-1)[0]
+        if self._aggr == "max":
+            x = x.squeeze().max(-1)[0]
+        else:
```
nicolas-chaulet: I suggest we follow the usual pattern here:
elif self._aggr == "mean" 
else raise NotImplemented

models/core_modules.py
```diff
@@ -78,11 +78,15 @@ def __init__(self, sampler, neighbour_finder, *args, **kwargs):
...
-            model_parameters = filter(lambda p: p.requires_grad, self.parameters())
-            self._params = sum([np.prod(p.size()) for p in model_parameters])
-        return self._params
+    def nb_params(self):
+        """[This property is used to return the number of trainable parameters for a given layer]
+        It is useful for debugging and reproducibility.
+        Returns:
+            [type] -- [description]
+        """
+        model_parameters = filter(lambda p: p.requires_grad, self.parameters())
```
nicolas-chaulet: You have removed the hastattr test by mistake I think

tchaton: I changed it, as it could be called when the model is in eval mode, and then used in train mode.
Like that, it is always recalculated.

PR #127, opened 2020-01-22
Decomposition hydra

tchaton:


tchaton:
@nicolas-chaulet, 

I changed it to create all models under all dataset to make it fault free.
try to fix create_all_models.

Best,
T.C

README.md
```diff
@@ -66,7 +66,7 @@ python -m unittest
 
 ## Train pointnet++ on part segmentation task for dataset shapenet
 ```
-poetry run python train.py experiment.model_name=pointnet2_charlesssg wandb.log=False experiment.data=shapenet
+poetry run python train.py task=segmentation model_type=pointnet2 model_name=pointnet2_charlesssg dataset=shapenet
```
omry: does model_type and model_name have a 1 to 1 mapping?
can model_name be a part of the model definition file?

conf/config.yaml
```diff
@@ -1,75 +1,26 @@
-# Those arguments within experiment defines which model, dataset and task to be created for benchmarking
-experiment:
-    model_name: pointnet2_charlesmsg
-    checkpoint_dir: ""
-    resume: True
-    task: segmentation
-    dataset: shapenet
+defaults:
+  - task: ??? # Task performed (segmentation, classification etc...)
```
omry: are all those things really optional?
I did not consider the combination of missing ('???') and optional.

conf/config.yaml
```diff
@@ -1,75 +1,26 @@
...
+    optional: True
+  - model_type: ??? # Type of model to use, e.g. pointnet2, rsconv etc...
+    optional: True
+  - dataset: ???
+    optional: True
+
+  - training
+
+  - models: ${defaults.0.task}/${defaults.1.model_type}
```
omry: Does the model depend on the task?
eg: do you have a different pointnet model for classification and segmentation?

consider a directory like:
```
conf/
  segmentation/model/
     model1.ymal
     model2.yaml
```

then arguments can be:
```
$ python foo.py segmentation/model=pointnet1
                 ^^^^^^^^^^^^^^^^^|^^^^^^^^^
               config group name  | choice in group
```

conf/config.yaml
```diff
@@ -1,75 +1,26 @@
...
+  - data: ${defaults.0.task}/${defaults.2.dataset}
+  - sota # Contains current SOTA results on different datasets (extracted from papers !).
+  - hydra/job_logging: custom
 
+model_name: ??? # Name of the specific model to load
```
omry: so this is a filename, right?
probably better to call it model_filename (or checkpoint_filename).

conf/data/classification/modelnet.yaml
```diff
@@ -0,0 +1,5 @@
+data:
+    task: classification
```
omry: why does the data need to know about the task?

README.md
nicolas-chaulet: No, model_type is a family of models and model_name is the specific model within that family. We have decided to keep one file per family as opposed to one file per model. It is nice to have other models in the same file just to be able to have examples. I know it's not "hydra" but I think it is a bit more user friendly

conf/config.yaml
nicolas-chaulet: They are not, but since there is no task/segmentation.yaml file we had to make them optional.

conf/config.yaml
nicolas-chaulet: Yes it does. But I guess the solution you suggest would work as well

conf/config.yaml
nicolas-chaulet: It's not a filename, it is a key within the model_type file. In each model file we have multiple configs (see my comment above) and model_name is used to resolve the correct one at run time.

conf/data/classification/modelnet.yaml
nicolas-chaulet: Yes, it should not. But it actually does in a way since depending on the task the data gets loaded differently. And we need to know which task is running at runtime to load the correct model (for example `src.models.segmentation.pointnet2.Model`). And instead  of giving the full module name we simply give the class within the specific task (pointnet2.Model). The rest of the path is resolved based on the task. The same happens for the datasets.

README.md
omry: I see.
As I suggested above you can have two level hierarchy to group models in each family together.
this can simply the code because it will no longer need to select which model in the family is the right one at runtime, while still making it clear that a model has a family and a type.

```
config/model/
  pointnet/
     pointnet2_charlesssg.yaml
     pointnet.yaml
```
etc

conf/data/classification/modelnet.yaml
omry: 1. consider separating into two families of data loaders.
2. in general when you have dependencies like this, you might be able to express them with interpolation:

```yaml
task:
   name: classification

data:
   task_name: ${task.name}
```

This way when you compose a task in, it will automatically reflect in in the composed data.

PR #129, opened 2020-01-24
Kpconvtest

nicolas-chaulet:
This PR contains the basics of a grid sampling enabled KPConv. 
The grid sampling happens on the GPU as well as the neighbourhood search. The current block SimpleBlock supports a strided (grid_size > 0) and a non strided (grid_size=0) mode.
Next steps are to add Resnet blocks and bottlenecks blocks


src/modules/KPConv/blocks.py
```diff
@@ -0,0 +1,65 @@
...
+from src.utils.enums import ConvolutionFormat
+
+
+class SimpleBlock(torch.nn.Module):
+    """
+    simple layer with KPConv convolution -> activation -> BN
+    we can perform a stride version (just change the query and the neighbors)
+    """
+
+    CONV_TYPE = ConvolutionFormat.PARTIAL_DENSE.value[-1]
```
tchaton: Conv_type is not needed if there are no check wit kwargs conv_type

PR #135, opened 2020-01-27
Kpconvresnet

nicolas-chaulet:
Still todo:

- [x]  validate the ResnetBlocks in the different modes (stride, bottleneck etc...)
- [x] Write a foward
- [ ] Add the gridsampling option to the datasets


conf/models/segmentation/kpconv.yaml
```diff
@@ -116,3 +117,66 @@ models:
...
+        architecture: kpconv.KPConvPaper
+        conv_type: "PARTIAL_DENSE"
+        define_constants:
+            in_grid_size: 0.02
+            in_feat: 64
+        down_conv:
+            down_conv_nn:
+                [
+                    [[FEAT + 1, in_feat], [in_feat, 2*in_feat]],
+                    [[2*in_feat, 2*in_feat], [2*in_feat, 4*in_feat]],
```
tchaton: Why do you have FEAT + 1 ?

src/datasets/base_dataset.py
```diff
@@ -30,9 +30,7 @@ def _create_dataloaders(self, train_dataset, test_dataset, val_dataset=None):
         self._feature_dimension = train_dataset.num_features
         if self._torch_loader:
             dataloader = partial(
-                torch.utils.data.DataLoader,
-                pin_memory=True,
-                collate_fn=lambda data_list: SimpleBatch.from_data_list(data_list),
+                torch.utils.data.DataLoader, collate_fn=lambda data_list: SimpleBatch.from_data_list(data_list),
```
tchaton: Are you sure you want to remove pin_memory ?

src/models/base_architectures/unet.py
```diff
@@ -377,44 +369,28 @@ def _init_from_compact_format(self, opt, model_type, dataset, modules_lib):
             down_conv_cls_name, up_conv_cls_name, modules_lib
         )  # Create the factory object
 
-        # construct unet structure
+        # Lobal module
```
tchaton: typo

src/models/base_architectures/unet.py
```diff
@@ -443,18 +419,18 @@ def _fetch_arguments_from_list(self, opt, index):
...
-        args_up["index"] = index
-        args_up["up_conv_cls"] = self._factory_module.get_module(index, "UP")
-        return args_up, args_down
+        Arguments:
+            conv_opt
+            index in sequential order (as they come in the config)
+            flow "UP" or "DOWN"
+        """
+        args = self._fetch_arguments_from_list(conv_opt, index)
+        args["conv_cls"] = self._factory_module.get_module(flow)
```
tchaton: Are we sure we don t want to provide the index to factory get_module function ?

src/models/segmentation/kpconv.py
```diff
@@ -1,6 +1,105 @@
...
+            if not dataset.class_to_segments:
+                raise ValueError(
+                    "The dataset needs to specify a class_to_segments property when using category information for segmentation"
+                )
+            self._num_categories = len(dataset.class_to_segments.keys())
+            log.info("Using category information for the predictions with %i categories", self._num_categories)
+        else:
+            self._num_categories = 0
+
+        # ASsemble encoder / decoder
```
tchaton: typo

src/models/segmentation/kpconv.py
```diff
@@ -1,6 +1,105 @@
...
+        self.loss_names = ["loss_seg"]
+
+    def set_input(self, data):
+        """Unpack input data from the dataloader and perform necessary pre-processing steps.
+        Parameters:
+            input: a dictionary that contains the data itself and its metadata information.
+        """
+        if hasattr(data, "x"):
+            ones = torch.ones(data.x.shape[0], dtype=torch.float).unsqueeze(-1).to(data.x.device)
+            data.x = torch.cat([ones, data.x], dim=-1)
```
tchaton: Why do you add ones ?

src/models/segmentation/kpconv.py
```diff
@@ -1,6 +1,105 @@
...
+        self.input = data
+        self.labels = data.y
+        self.batch_idx = data.batch
+
+    def forward(self) -> Any:
+        """Run forward pass. This will be called by both functions <optimize_parameters> and <test>."""
+        stack_down = []
+
+        data = self.input
+        for i in range(len(self.down_modules) - 1):
```
tchaton: Much cleaner forward function now ;)

src/models/segmentation/kpconv.py
```diff
@@ -1,6 +1,105 @@
...
+        last_feature = self.FC_layer(last_feature)
+        self.output = F.log_softmax(last_feature, dim=-1)
+        return self.output
+
+    def backward(self):
+        """Calculate losses, gradients, and update network weights; called in every training iteration"""
+        # caculate the intermediate results if necessary; here self.output has been computed during function <forward>
+        # calculate loss given the input and intermediate results
+
+        self.loss_seg = F.nll_loss(self.output, self.labels) + self.get_internal_loss()
```
tchaton: You might want to use self._weight_classes

conf/models/segmentation/kpconv.yaml
```diff
@@ -116,3 +117,66 @@ models:
...
+            module_name: FPModule
+            up_conv_nn:
+                [
+                    [32*in_feat + 16*in_feat, 8*in_feat],
+                    [8*in_feat + 8*in_feat, 4*in_feat],
+                    [4*in_feat + 4*in_feat, 2*in_feat],
+                    [2*in_feat + 2*in_feat, in_feat],
+                ]
+            skip: True
+            up_k: [1, 1, 1, 1]
```
tchaton: Are you sure it is 1, 1, 1, 1 ? Looks a bit wierd

src/modules/KPConv/blocks.py
```diff
@@ -17,32 +19,33 @@ class SimpleBlock(torch.nn.Module):
...
         self,
         down_conv_nn=None,
-        radius=None,
-        kp_extent=1,
-        density_parameter=1,
-        max_num_neighbors=64,
+        grid_size=None,
+        is_strided=False,
+        sigma=1.0,
+        density_parameter=2.5,
```
tchaton: is the density parameter always 2.5 ?

src/modules/KPConv/blocks.py
```diff
@@ -51,14 +54,158 @@ def forward(self, data):
             querry_data = data.clone()
 
         q_pos, q_batch = querry_data.pos, querry_data.batch
-        idx_neighbour, _ = self.neighbour_finder(q_pos, data.pos, batch_x=q_batch, batch_y=data.batch)
-        x = self.kp_conv(q_pos, data.pos, idx_neighbour, data.x,)
+        # if hasattr(data, "idx_neighboors") and data.idx_neighboors.shape[0] == q_pos.shape[0]:
+        #     idx_neighboors = data.idx_neighboors
+        # else:
+        idx_neighboors, _ = self.neighbour_finder(data.pos, q_pos, batch_x=data.batch, batch_y=q_batch)
```
tchaton: Is query and support points the same ? If yes, maybe clone should not be necessary here.

src/modules/KPConv/blocks.py
```diff
@@ -17,32 +19,33 @@ class SimpleBlock(torch.nn.Module):
...
+        grid_size=None,
+        is_strided=False,
+        sigma=1.0,
+        density_parameter=2.5,
+        max_num_neighbors=16,
         activation=torch.nn.LeakyReLU(negative_slope=0.2),
         bn_momentum=0.1,
         bn=torch.nn.BatchNorm1d,
-        grid_size=None,
         **kwargs
```
tchaton: Extract conv_type and make sure it is the same as CONV_TYPE

src/modules/KPConv/blocks.py
```diff
@@ -51,14 +54,158 @@ def forward(self, data):
...
+        grid_size=None,
+        is_strided=False,
+        sigma=1,
+        density_parameter=2.5,
+        max_num_neighbors=16,
+        activation=torch.nn.LeakyReLU(negative_slope=0.2),
+        has_bottleneck=True,
+        bn_momentum=0.1,
+        bn=torch.nn.BatchNorm1d,
+        **kwargs
```
tchaton: Same as before.

src/modules/KPConv/blocks.py
```diff
@@ -51,14 +54,158 @@ def forward(self, data):
...
+    ):
+        super(ResnetBBlock, self).__init__()
+        assert len(down_conv_nn) == 2 or len(down_conv_nn) == 3
+        if len(down_conv_nn) == 2:
```
tchaton: If you except to have len in [2, 3], then you should add an exception here, just in case.

src/modules/KPConv/blocks.py
```diff
@@ -51,14 +54,158 @@ def forward(self, data):
...
+            num_inputs, num_outputs = down_conv_nn
+            d_2 = num_outputs // 4
+        else:
+            num_inputs, d_2, num_outputs = down_conv_nn
+        self.is_strided = is_strided
+        self.grid_size = grid_size
+        self.has_bottleneck = has_bottleneck
+        if self.is_strided:
+            self.sampler = GridSampling(grid_size)
```
tchaton: How comes the sampler is never used ?

src/modules/KPConv/blocks.py
```diff
@@ -51,14 +54,158 @@ def forward(self, data):
...
 
         querry_data.x = x
+        querry_data.idx_neighboors = idx_neighboors
         return querry_data
 
     def extra_repr(self):
         return str(self.sampler) + "," + str(self.neighbour_finder)
+
+
+class ResnetBBlock(torch.nn.Module):
```
tchaton: If you make them inherit from nn.Module, maybe you could use the BaseConvolution instead.

src/modules/KPConv/blocks.py
```diff
@@ -51,14 +54,158 @@ def forward(self, data):
...
+class KPDualBlock(torch.nn.Module):
+    def __init__(
+        self, block_names=None, down_conv_nn=None, grid_size=None, is_strided=None, has_bottleneck=None, **kwargs
+    ):
+        super(KPDualBlock, self).__init__()
+
+        assert len(block_names) == len(down_conv_nn)
+        self.blocks = torch.nn.ModuleList()
+        for i, class_name in enumerate(block_names):
+            kpcls = getattr(sys.modules["src.modules.KPConv.blocks"], class_name)
```
tchaton: Not sure I really like having direct path there. What happen if we give a proper name to src ?

src/models/segmentation/kpconv.py
```diff
@@ -1,6 +1,105 @@
...
+        for i in range(len(self.up_modules)):
+            data = self.up_modules[i]((data, stack_down.pop()))
+
+        last_feature = data.x
+        if self._use_category:
+            num_points = data.pos.shape[1]
+            cat_one_hot = (
+                torch.zeros((data.pos.shape[0], self._num_categories, num_points)).float().to(self.category.device)
+            )
+            cat_one_hot.scatter_(1, self.category.repeat(1, num_points).unsqueeze(1), 1)
```
tchaton: You could use F.one_hot function to do one hot encoding

tchaton:
Excellent PR ! Lot of great progress for the framework. 
I think GridSampling could be the bottleneck due to scatter_mean used to aggregate points.
If we manage to compile torch_scatter 1.5.0, we could use segment_mean, which would be much faster.

Best,
T.C

conf/models/segmentation/kpconv.yaml
```diff
@@ -116,3 +117,66 @@ models:
...
+    KPConvPaper:
+        architecture: kpconv.KPConvPaper
+        conv_type: "PARTIAL_DENSE"
+        define_constants:
+            in_grid_size: 0.02
+            in_feat: 64
+        down_conv:
+            down_conv_nn:
+                [
+                    [[FEAT + 1, in_feat], [in_feat, in_feat]],
```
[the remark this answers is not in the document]
nicolas-chaulet: As I understand KPConv, they go in pairs. I thought it would be nice to make it explicit. The first is the strided the second channel rising

conf/models/segmentation/kpconv.yaml
```diff
@@ -116,3 +117,66 @@ models:
...
+                [
+                    ["SimpleBlock", "ResnetBBlock"],
+                    ["ResnetBBlock", "ResnetBBlock"],
+                    ["ResnetBBlock", "ResnetBBlock"],
+                    ["ResnetBBlock", "ResnetBBlock"],
+                    ["ResnetBBlock", "ResnetBBlock"],
+                ]
+            has_bottleneck:
+                [
+                    [False, False],
```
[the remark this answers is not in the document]
nicolas-chaulet: Good spot!

src/models/segmentation/kpconv.py
nicolas-chaulet: This is actually never hit at the moment. Fixing it

nicolas-chaulet: @tchaton I have changed the way categories are stored in the preprocessed data. You will have to re process shapenet

src/modules/KPConv/blocks.py
nicolas-chaulet: Good point. Especially since __name__ does exactly that

src/modules/KPConv/blocks.py
nicolas-chaulet: It is used in the first line of forward

src/modules/KPConv/blocks.py
nicolas-chaulet: There is an assertion that would throw an exception. I have added a message

conf/models/segmentation/kpconv.yaml
nicolas-chaulet: This is how KPConv features work, 1 is added to the features

src/datasets/base_dataset.py
nicolas-chaulet: It did not change anything and going through pytorch blogs it may even be dangerous. So I removed it for now but can add it back if you want

src/models/base_architectures/unet.py
nicolas-chaulet: This is never used anywhere at the moment. So I removed it. We can add it back when we need it

conf/models/segmentation/kpconv.yaml
nicolas-chaulet: Get closest nearest neighbour. It is like that in the official code base I think

src/modules/KPConv/blocks.py
nicolas-chaulet: It would be 5 for deformable, but yes, all datasets share the same, it is dataset agnostic

src/models/segmentation/kpconv.py
nicolas-chaulet: Done

src/modules/KPConv/blocks.py
nicolas-chaulet: Good shout!

src/modules/KPConv/blocks.py
nicolas-chaulet: They would need a sampler and a neighbourhood finder which they don't sometimes (ResnetBlock does not for example)

nicolas-chaulet: Not sure it makes much sense here... inheriting the number of parameters function would be nice, but the neighbourhood finder and sampler does not make so much sense at the moment (sometimes they don't have one)

src/modules/KPConv/blocks.py
nicolas-chaulet: I actually removed the checks some time ago because it was cluttering the config. You have a good point though, if you use the wrong convolution module with the wrong architectures things will break in a weird way. I think a nicer way to handle that would be in the unet or the main model, when it creates itself it makes sure that blocks it uses are compatible. Let's log that and revisit

nicolas-chaulet:
Thanks for the review @tchaton ! It is much better now

PR #136, opened 2020-01-28
Add validation and associated elements

tchaton:


eval.py
```diff
@@ -0,0 +1,131 @@
...
+from src.metrics.colored_tqdm import Coloredtqdm as Ctq
+from src.metrics.model_checkpoint import get_model_checkpoint, ModelCheckpoint
+
+# Utils import
+from src.utils.model_building_utils.model_definition_resolver import resolve_model
+from src.utils.colors import COLORS
+from src.utils.config import merges_in_sub, set_format
+
+
+def enable_dropout_in_eval(model):
```
nicolas-chaulet: Could you please make it a member of the base_model?
Something like `activate_dropout()`, it will be a bit cleaner.

eval.py
```diff
@@ -0,0 +1,131 @@
...
+    # Find and create associated model
+    resolve_model(model_config, dataset, tested_task)
+    model_config = merges_in_sub(model_config, [cfg_eval, dataset_config])
+    model = find_model_using_name(model_config.architecture, tested_task, model_config, dataset)
+
+    log.info(model)
+
+    # 
+    if cfg_eval.enable_dropout:
+        enable_dropout_in_eval(model)
```
nicolas-chaulet: This would change a bit with the suggest update above.

PR #141, opened 2020-01-31
update documentation

tchaton:


tchaton:
@nicolas-chaulet, no I didn't. It is because of the gitignore. I will add the back.

PR #142, opened 2020-01-31
Gridsampling

nicolas-chaulet:
Still need to update the other datasets but I think it is for the better


conf/data/segmentation/shapenet.yaml
```diff
@@ -3,5 +3,10 @@ data:
...
     dataroot: data
     normal: True
-    num_points: 2048
     first_subsampling: 0.02
+    num_points: 2048
+    pre_transforms:
+        - transform: NormalizeScale
+        - transform: GridSampling
+          params:
+              size: ${data.first_subsampling}
```
tchaton: Great finding !

src/core/data_transform/__init__.py
```diff
@@ -1 +1,53 @@
+import sys
+
+import torch_geometric.transforms as T
 from .transforms import *
+
+_custom_transforms = sys.modules[__name__]
+_torch_geometric_transforms = sys.modules["torch_geometric.transforms"]
+_intersection = set(_custom_transforms.__dict__) & set(_torch_geometric_transforms.__dict__)
+_intersection = [module for module in _intersection if not module.startswith("_")]
+if _intersection:
```
tchaton: Great idea !

tchaton:
Great PR so far ! Do you plan to extend it for train, test, val transform too in this PR ?

nicolas-chaulet:
@tchaton I'll do the transforms in another PR. Let's try with this setting and if all works fine then let's use it across the board

PR #143, opened 2020-02-03
add suport for file_class_pointer

tchaton:


tchaton:
@nicolas-chaulet, KPConv tests are breaking. Can you have a look ?

src/__init__.py
```diff
@@ -11,18 +11,19 @@ def contains_key(opt, key):
         return False
 
 
-def find_dataset_using_name(dataset_name, tested_task):
-    """Import the module "data/[dataset_name]_dataset.py".
-    In the file, the class called DatasetNameDataset() will
+def find_dataset_using_name(tested_dataset_file_class_pointer, tested_task):
+    """Import the module "data/[dataset_file_name].py".
```
nicolas-chaulet: data/task/datasetfilename.py

src/datasets/segmentation/s3dis.py
```diff
@@ -234,52 +235,126 @@ def process(self):
...
 
-"""
-class NormalizeMeanStd(object):
 
-    def __init__(self, keys):
-        self._keys = keys
+class S3DISDataset(BaseDataset):
+    def __init__(self, dataset_opt, training_opt):
+        super().__init__(dataset_opt, training_opt)
+        self._data_path = os.path.join(dataset_opt.dataroot, "S3DIS")
```
nicolas-chaulet: I think it is the right PR to change that as well. Remove the `Dataset` postfix from the classname in all datasets (same for the filenames) and use the classname as the data_path

src/datasets/segmentation/s3dis.py
```diff
@@ -234,52 +235,126 @@ def process(self):
         torch.save(self.collate(train_data_list), self.processed_paths[0])
         torch.save(self.collate(test_data_list), self.processed_paths[1])
 
+class S3DISOriginal_With_Weights(S3DISOriginal):
```
nicolas-chaulet: Why would we have one with and without weights? I would suggest we only have one with weights.

src/datasets/segmentation/s3dis.py
```diff
@@ -234,52 +235,126 @@ def process(self):
...
+            dataset {[type]}
+            wandb_log - Log using weight and biases
+        Returns:
+            [BaseTracker] -- tracker
+        """
+        return SegmentationTracker(dataset, wandb_log=wandb_opt.log, use_tensorboard=tensorboard_opt.log)
 
 
-class S3DIS_With_Weights(S3DIS):
+class S3DIS_pyg_With_Weights(S3DIS_pyg):
```
nicolas-chaulet: Why is it called pyg? I guess the 1x1 notation was a bit more explicit

conf/data/classification/modelnet.yaml
```diff
@@ -1,5 +1,6 @@
 data:
     task: classification
+    file_class_pointer: modelnet.ModelNetDataset
```
nicolas-chaulet: file_class_pointer is a bit too verbose. Something like `class` or `class_name` would be shorter.

src/__init__.py
```diff
@@ -11,18 +11,19 @@ def contains_key(opt, key):
         return False
 
 
-def find_dataset_using_name(dataset_name, tested_task):
-    """Import the module "data/[dataset_name]_dataset.py".
-    In the file, the class called DatasetNameDataset() will
+def find_dataset_using_name(tested_dataset_file_class_pointer, tested_task):
```
nicolas-chaulet: I would suggest renaming the parameters and method to `instantiate_dataset(class_name, task)`

train.py
```diff
@@ -132,9 +134,9 @@ def main(cfg):
 
     # Find and create associated dataset
     dataset_config = cfg.data
-    tested_dataset_name = dataset_config.name
+    tested_dataset_file_class_pointer = dataset_config.file_class_pointer
```
nicolas-chaulet: rename to `dataset_class`

tchaton: I renamed to class

PR #145, opened 2020-02-04
Precompute radius search and grid sampling on GPU

nicolas-chaulet:
THis PR contains the code for building the grid sampling and neighbours on the CPU during the data loading process. This produces a speed up of 4x on Shapenet training for KPConv.


src/core/data_transform/transforms.py
```diff
@@ -183,34 +182,36 @@ def new__inc__(key, num_nodes, special_params=None, func=None):
...
+        # Compute sequentially multi_scale indexes on cpu
+        ms_data = MultiScaleData.from_data(data)
+        precomputed = [data]
+        for index in range(self.num_layers):
+            sampler, neighbour_finder = self.strategies["sampler"][index], self.strategies["neighbour_finder"][index]
+            support = precomputed[index]
+            if sampler:
+                query = sampler(support.clone())
+            else:
+                query = support.clone()
```
tchaton: Need to check if clone is necessary here.

src/datasets/base_dataset.py
```diff
@@ -153,9 +167,9 @@ def _set_multiscale_transform(self, transform):
                             attr.dataset, "transform", Compose([current_transform, transform]),
                         )
 
-    def set_strategies(self, model, precompute_multi_scale=False):
+    def set_strategies(self, model):
         strategies = model.get_sampling_and_search_strategies()
```
tchaton: It won't work with model which aren't implementing UNetBaseModel

src/datasets/multiscale_data.py
```diff
@@ -0,0 +1,76 @@
+from typing import List, Optional
+import torch
+import copy
+import torch_geometric
+from torch_geometric.data import Data
+from torch_geometric.data import Batch
+
+
+class MultiScaleData(Data):
```
tchaton: I think we should rename to MultiResolutionData, and MultiScale should be if several radiuses are provided.

src/datasets/multiscale_data.py
```diff
@@ -0,0 +1,76 @@
...
+    def from_data(cls, data):
+        ms_data = cls()
+        for k, item in data:
+            ms_data[k] = item
+        return ms_data
+
+
+class MultiScaleBatch(MultiScaleData):
+    @staticmethod
+    def from_data_list(data_list, follow_batch=[]):
```
tchaton: Great !

src/modules/KPConv/blocks.py
```diff
@@ -47,25 +47,34 @@ def __init__(
         else:
             self.sampler = None
 
-    def forward(self, data):
-        if self.sampler:
-            querry_data = self.sampler(data.clone())
+    def forward(self, data, pre_computed=None, **kwargs):
+        if not hasattr(data, "block_idx"):
+            setattr(data, "block_idx", 0)
```
tchaton: Why do you set setattr(data, "block_idx", 0) ?

tchaton:
Great PR ! Small things to change and should be good to go.

src/modules/KPConv/blocks.py
tchaton: Not sure it is good idea. Let's switch with to dictionnary based pre_computed.

src/datasets/base_dataset.py
nicolas-chaulet: No, but I guess that's fine for now

src/core/data_transform/transforms.py
nicolas-chaulet: Yes it is, otherwise in the forward methods that would use those precomputed data there is a risk that it overrides a previous iteration (for example its features) which would break things

src/datasets/multiscale_data.py
```diff
@@ -0,0 +1,133 @@
...
+    for key in follow_batch:
+        batch["{}_batch".format(key)] = []
+
+    cumsum = {key: 0 for key in keys}
+    batch.batch = []
+    for i, data in enumerate(data_list):
+        for key in data.keys:
+            item = data[key]
+            if torch.is_tensor(item) and item.dtype != torch.bool:
+                mask = item >= 0
```
tchaton: Do you see any scenario where this mask could be a trouble ?

PR #146, opened 2020-02-05
Descriptors/dataset

humanpose1:
The whole pipeline for processing 3D match to get fragments.
It downloads the dataset (it is heavy), it contains RGBD frames and the pose and the intrinsic params
Then it creates fragment using [this code](https://github.com/andyzeng/tsdf-fusion-python).
Then it precomputes matches between two fragments (if the overlap is bigger than a parameter).

Still things to do:

- [x] -Hydra conf file.
- [x] Add a Fragment Dataset to train like FCGF (Maybe later).
- [x]  Add a tracker
- [x] pointnet example
- [ ]  dataset for testing
- [ ] saving features
- [ ] script to compute recall
- [ ] script to registrate scene using RANSAC


src/datasets/registration/fragment3dmatch.py
```diff
@@ -0,0 +1,200 @@
...
+from torch_geometric.data import Dataset, Data, download_url, extract_zip
+from torch_geometric.data import DataLoader
+
+from src.datasets.base_dataset import BaseDataset
+from src.datasets.registration.utils import rgbd2fragment
+
+log = logging.getLogger(__name__)
+
+
+def to_list(x):
```
tchaton: Those helpers should be put within the utils, if there isn't duplicates.

src/datasets/registration/fragment3dmatch.py
```diff
@@ -0,0 +1,200 @@
...
+    res = []
+    with open(filename, 'r') as f:
+        res = f.readlines()
+    return res
+
+
+class General3DMatch(Dataset):
+
+    list_urls_train = get_urls(osp.join('urls', 'url_train.txt'))
+    list_urls_train_small = get_urls(osp.join('urls', 'url_train_small.txt'))
```
tchaton: Maybe url_{}.txt could be automatically downloaded.

src/datasets/registration/fragment3dmatch.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def get_urls(filename):
+    res = []
+    with open(filename, 'r') as f:
+        res = f.readlines()
+    return res
+
+
+class General3DMatch(Dataset):
```
tchaton: Maybe, all 3DMatch dataset should be within the same file, using the new API for dataset loading.

class: {file_name}.{dataset_class_name}

src/datasets/registration/patch3dmatch.py
```diff
@@ -0,0 +1,133 @@
...
+
+        if(self.transform is not None):
+            data_source = self.transform(data_source)
+            data_target = self.transform(data_target)
+        batch = Batch.from_data_list([data_source, data_target])
+        batch.pair = batch.batch
+        batch.batch = None
+        return batch
+
+class Patch3DMatchDataset(BaseDataset):
```
tchaton: Great !

src/datasets/registration/fragment3dmatch.py
tchaton: Instead of being part of the project. @nicolas-chaulet, what are you thoughts ?

src/datasets/registration/fragment3dmatch.py
humanpose1: I will change the name of the file and adopt the convention.
General3DMatch must not be used directly, that's why it hasn't a dataloader.
It is just a class that download and process the data.
Patch3DMatch or Fragment3DMatch must be used instead.

src/datasets/registration/patch3dmatch.py
```diff
@@ -0,0 +1,174 @@
...
+            radius_patch=dataset_opt.radius_patch,
+            num_frame_per_fragment=dataset_opt.num_frame_per_fragment,
+            max_dist_overlap=dataset_opt.max_dist_overlap,
+            min_overlap_ratio=dataset_opt.min_overlap_ratio,
+            tsdf_voxel_size=dataset_opt.tsdf_voxel_size,
+            depth_thresh=dataset_opt.depth_thresh,
+            pre_transform=pre_transform,
+            transform=test_transform)
+
+        self._create_dataloaders(train_dataset, test_dataset)
```
nicolas-chaulet: We are in the process of changing the way datasets and dataloader get created -> See #166. You now simply need to set a `test_dataset` and `train_dataset` members to your dataset object. The train.py scripts takes care of creating the data loaders

src/datasets/registration/patch3dmatch.py
```diff
@@ -0,0 +1,174 @@
...
+        Factory method for the tracker
+
+        Arguments:
+            task {str} -- task description
+            dataset {[type]}
+            wandb_log - Log using weight and biases
+        Returns:
+            [BaseTracker] -- tracker
+        """
+        return PatchRegistrationTracker(dataset, task, wandb_opt.log,
```
nicolas-chaulet: This has changed as well. The wandb and tensorboard options are booleans

src/datasets/registration/utils.py
```diff
@@ -0,0 +1,276 @@
...
+        pcd, color = extract_pcd(depth, intrinsic, color_image)
+        pcd = pcd.dot(trans[:3, :3].T) + trans[:3, 3]
+        return pcd, color
+    else:
+        pcd = extract_pcd(depth, intrinsic)
+        pcd = pcd.dot(trans[:3, :3].T) + trans[:3, 3]
+        return pcd
+
+
+def rgbd2fragment_rough(list_path_img,
```
nicolas-chaulet: Are those functions yours? If not please reference the source.

src/metrics/registration_accuracy.py
```diff
@@ -0,0 +1,18 @@
+import numpy as np
+from sklearn.neighbors import NearestNeighbors
+
+
+def compute_accuracy(embedded_ref_features, embedded_val_features):
+
+    """
+    evaluate accuracy for a batch
```
nicolas-chaulet: Could you please document the input? What are they supposed to contain?

src/metrics/registration_tracker.py
```diff
@@ -0,0 +1,42 @@
...
+        self._acc = compute_accuracy(outputs[::2], outputs[1::2])
+        loss_dict = model.get_current_losses()
+        self._loss = loss_dict["loss"]
+        self._loss_reg = loss_dict["loss_reg"]
+
+    def get_metrics(self, verbose=False) -> Dict[str, float]:
+        """ Returns a dictionnary of all metrics and losses being tracked
+        """
+        metrics = super().get_metrics(verbose)
+        metrics["{}_loss".format(self._stage)] = self._loss
```
nicolas-chaulet: You don't need to add the losses, they are part of the `super()` call

src/models/base_architectures/backbone.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    create a backbone-based generator:
+    This is simply an encoder
+    (can be used in classification, regression, metric learning and so one)
+    """
+
+    def _save_sampling_and_search(self, down_conv):
+        sampler = getattr(down_conv, "sampler", None)
+        if is_list(sampler):
+            self._sampling_and_search_dict["sampler"] += sampler
```
nicolas-chaulet: `_sampling_and_search_dict` has been renamed to `_spatial_ops_dict`. Refer to teh Unet as an example

src/datasets/registration/fusion.py
```diff
@@ -0,0 +1,392 @@
+# Copyright (c) 2018 Andy Zeng
+# Taken from https://github.com/andyzeng/tsdf-fusion-python
+
+import numpy as np
+
+from numba import njit, prange
```
nicolas-chaulet: Could you please add numba to the poetry dependencies? (poetry add numba)

src/datasets/registration/utils.py
humanpose1: rgbd2fragment_rough is mine (simply concatenate point cloud frame rgbd frame without intelligent fusion). but the fragment are not beautiful with this function.
for rgbd2fragment_fine, I use the code from andy zeng (I have adapted it a bit).

src/core/losses/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .losses import *
+from pytorch_metric_learning.miners import *
```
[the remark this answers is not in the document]
humanpose1: I wanted to do something like instantiate_transforms for the losses and the miners so the user can directly choose the loss from the yaml.
Now it's normally cleaner than before

src/datasets/registration/fusion.py
humanpose1: I'm sorry. I've had some problems with poetry. Actually, I'm simply using an anaconda env and I've installed the package manually.
We need to add numba, pycuda and pytorch_metric_learning.

src/datasets/registration/general3dmatch_dataset.py
```diff
@@ -0,0 +1,229 @@
+import logging
```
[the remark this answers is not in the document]
humanpose1: I can download a list in the website of 3dmatch [here](http://3dmatch.cs.princeton.edu/), they even have a bash script to download the dataset. But the training dataset and the testing dataset are not split (they also have a split.txt). 
I have made the url.txt manually...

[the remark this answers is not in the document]
humanpose1: also, the training dataset and the testing dataset are different. In the test dataset, we must download precomputed fragments (to have the same data as other papers for the testing).

src/datasets/registration/fusion.py
tchaton: Did you do that : https://deeppointcloud-benchmarks.readthedocs.io/en/latest/src/installation/

src/datasets/registration/general3dmatch_dataset.py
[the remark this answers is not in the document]
nicolas-chaulet: So, if I understand correctly you'd rather keep the txt file as is in the repo?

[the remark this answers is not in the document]
humanpose1: For now, yes even if it's not very elegant... but I will solve this problem later.

PR #147, opened 2020-02-05
add new transform

tchaton:


src/datasets/segmentation/s3dis.py
```diff
@@ -173,6 +160,21 @@ def __init__(
...
+        self.kd_trees = self.load_kd_trees("train" if train else "test")
+
+    def __getitem__(self, idx):
+        r"""Gets the data object at index :obj:`idx` and transforms it (in case
+        a :obj:`self.transform` is given).
+        In case :obj:`idx` is a slicing object, *e.g.*, :obj:`[2:5]`, a list, a
+        tuple, a  LongTensor or a BoolTensor, will return a subset of the
+        dataset at the specified indices."""
+        if isinstance(idx, int):
+            data = self.get(idx)
```
nicolas-chaulet: Are you sure that the order is preserved? It feels like you could use something like the room name (they are unique) to save the trees. The name would be stored in the data object as well, like that you have a unique key to access your trees. No risk of indexes being shuffled

conf/models/segmentation/kpconv.yaml
```diff
@@ -121,7 +121,7 @@ models:
     KPConvPaper:
         class: kpconv.KPConvPaper
         conv_type: "PARTIAL_DENSE"
-        use_category: True
+        use_category: False
```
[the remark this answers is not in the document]
tchaton: Great idea !

conf/models/segmentation/kpconv.yaml
```diff
@@ -121,8 +121,8 @@ models:
     KPConvPaper:
         class: kpconv.KPConvPaper
         conv_type: "PARTIAL_DENSE"
-        use_category: True
         define_constants:
+            use_category: ${data.use_category}
```
nicolas-chaulet: I don't think this should be in `define_sontants`. Also, could you change it everywhere in the configs? Just to be consistent?

src/datasets/segmentation/s3dis.py
```diff
@@ -222,8 +244,9 @@ def process(self):
         ]
 
         train_data_list, test_data_list = [], []
-
-        for (area, room_name, file_path) in tq(train_files + test_files):
+        
+        data_count = 0
+        for (area, room_name, file_path) in tq(train_files[:40] + test_files[:2]):
```
nicolas-chaulet: Remove this 40

conf/models/segmentation/kpconv.yaml
tchaton: It didn t work outside of it

PR #148, opened 2020-02-06
Kpconvdetails

nicolas-chaulet:
This PR contains some adjustements to KPConv to match Hugues' implementation a bit closer. 

- Random anisotropic scaling
- Introduction of gradient clip
- Switching optimizer to SGD + more flexibility in defining them
- bn momentum as a parameter in up convs


src/datasets/segmentation/shapenet.py
```diff
@@ -198,7 +198,7 @@ def __init__(self, dataset_opt, training_opt):
         except KeyError:
             self._category = None
         pre_transform = self._pre_transform
-        train_transform = T.Compose([RandomNoise()])
+        train_transform = T.Compose([RandomNoise(sigma=0.001, clip=0.005), RandomScaleAnisotropic((0.8, 1.2))])
```
tchaton: Could do put those transform within the configuration file instead ?

nicolas-chaulet:
Yes as soon as your PR for s3dis is merged. It contains the logic to handle
the transforms in the dataset config

tchaton:
@nicolas-chaulet , can you put the transform within the configuration ?

PR #149, opened 2020-02-07
fused dataset

tchaton:


conf/models/segmentation/pointnet2.yaml
```diff
@@ -86,7 +86,8 @@ models:
     pointnet2_onehot:
         class: pointnet2.PointNet2_D
         conv_type: "DENSE"
-        use_category: True
+        define_constants:
```
nicolas-chaulet: This isn't enough, you also need to define a `use_category` key in the pointnet2_onehot object. Otherwise the model will not see it. (Same goes for the other models)

src/core/data_transform/transforms.py
```diff
@@ -14,8 +16,91 @@
 
 from src.datasets.multiscale_data import MultiScaleData
 from src.utils.transform_utils import SamplingStrategy
+from src.utils.config import is_list
+from torch_geometric.data import Data, Batch
+from tqdm import tqdm as tq
 from src.utils import is_iterable
 
+class PointCloudFusion(object):
```
nicolas-chaulet: A docstring would be nice here

src/core/data_transform/transforms.py
```diff
@@ -14,8 +16,91 @@
...
+        return data
+
+    def __repr__(self):
+        return "{}()".format(self.__class__.__name__)
+
+class GridSphereSampling(object):
+    r"""Fit the point cloud to a grid and for each point in this grid, 
+    create a sphere with a radius r
+    Args:
+        radius (float or [float] or Tensor): Radius of the sphere to be sampled.
```
nicolas-chaulet: Add the grid_size parameter and the transform returns (A list of data objects, one per sphere)

src/core/data_transform/transforms.py
nicolas-chaulet: If PointCloudFusion and GridSpehereSampling are pre collate transform then it might be a good idea to extract them in another file called `precollate_transforms.py`. This will reduce the risk of using those in the wrong way (as pre-transforms or transforms)

src/core/data_transform/transforms.py
```diff
@@ -14,8 +16,91 @@
...
+        # The kdtree has bee attached to data for optimization reason.
+        # However, it won't be used for down the transform pipeline and should be removed before any collate func call.
+        if hasattr(data, self.KDTREE_KEY) and self._delattr_kd_tree:
+            delattr(data, self.KDTREE_KEY)
+
+        # apply grid sampling
+        grid_data = self._grid_sampling(data.clone())
+
+        datas = []
+        for grid_center, grid_label in zip(np.asarray(grid_data.pos), np.asarray(grid_data.y)):
```
nicolas-chaulet: This logic will work only if the data object is not a batch. Could you add an assert to make sure that batch is not present in the data?

src/core/data_transform/transforms.py
```diff
@@ -38,14 +130,15 @@ class RandomSphere(object):
         radius (float or [float] or Tensor): Radius of the sphere to be sampled.
```
nicolas-chaulet: Can you please add a description of what this transform returns? Also explain the centre option

switch_cuda.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
nicolas-chaulet: Not sure we want this file in our repo. Why did you add it?

src/datasets/samplers.py
```diff
@@ -0,0 +1,28 @@
+import torch
+import numpy as np
+from torch.utils.data import Sampler
+
+class BalancedRandomSampler(Sampler):
+    def __init__(self, labels, replacement=True):
```
nicolas-chaulet: A docstring would be nice

src/datasets/segmentation/s3dis.py
```diff
@@ -131,6 +137,52 @@ def add_weights(dataset, train, class_weight_method):
...
 
+################################### 1m cylinder s3dis ###################################
+
+class S3DIS1x1Dataset(BaseDataset):
+    def __init__(self, dataset_opt, training_opt):
+        super().__init__(dataset_opt, training_opt)
+
+        pre_transform = self._pre_transform
+
+        transform = T.Compose(
```
nicolas-chaulet: Could you use the config file to set this transform?

src/datasets/segmentation/s3dis.py
```diff
@@ -326,33 +378,193 @@ def get_tracker(model, task: str, dataset, wandb_opt: bool, tensorboard_opt: boo
...
+    def extract_center_labels(self, data_list):
+        extract_center_labels = []
+        for data in data_list:
+            if hasattr(data, "center_label"):
+                extract_center_labels.append(getattr(data, "center_label"))
+                delattr(data, "center_label")
+        return extract_center_labels
+
+    def save_center_labels(self, split_name, center_labels):
+        name = "{}_center_labels.p".format(split_name)
```
nicolas-chaulet: Please use .pt as a pytorch extension to be consistent

src/datasets/segmentation/s3dis.py
```diff
@@ -326,33 +378,193 @@ def get_tracker(model, task: str, dataset, wandb_opt: bool, tensorboard_opt: boo
...
+               
+            torch.save(data_list, self.pre_processed_path)
+        else:
+            data_list = torch.load(self.pre_processed_path)
+
+        train_data_list = [data_list[i] for i in range(6) if i != self.test_area - 1]
+        test_data_list = data_list[self.test_area - 1]
+
+        if self.pre_collate_transform:
+            print('pre_collate_transform ...')
```
nicolas-chaulet: Use log instead of print please

src/core/data_transform/transforms.py
tchaton: I would prefer not too. They don't need to be used at this stage, and could be used as pre_transform.

tchaton: I would prefer them not to be separated.

conf/models/segmentation/rsconv.yaml
```diff
@@ -57,7 +57,8 @@ models:
     RSConv_MSN:
         class: rsconv.RSConvLogicModel
         conv_type: "DENSE"
-        use_category: True
+        define_constants:
```
nicolas-chaulet: My point was that you need to define a `use_category` member inside each model. Otherwise it gets missed.

src/datasets/segmentation/s3dis.py
nicolas-chaulet: transform is not used now, could you please remove it?

PR #151, opened 2020-02-10
Loss computation and transforms

nicolas-chaulet:
This PR fixes a major bug in the evaluation of the loss on the test set. It also moves the logic for creating transforms out of the code and in the config file for shapenet

### DIFF
--- a/conf/data/segmentation/shapenet-fixed.yaml
+++ b/conf/data/segmentation/shapenet-fixed.yaml
@@ -1,0 +1,17 @@
+data:
+    class: shapenet.ShapeNetDataset
+    task: segmentation
+    dataroot: data
+    normal: True
+    train_size: 14016 # Number of shapes in the whole training set
+    use_category: True
+    train_transforms:
+        - transform: FixedPoints
+          lparams: [2048]
+        - transform: RandomNoise
+          params:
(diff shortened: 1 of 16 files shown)


conf/data/segmentation/shapenet-fixed.yaml
```diff
@@ -0,0 +1,23 @@
...
+    use_category: True
+    first_subsampling: 0.02
+    pre_transforms:
+      - transform: NormalizeScale
+      - transform: GridSampling
+        params:
+            size: ${data.first_subsampling}
+    train_transforms:
+        - transform: FixedPoints
+          lparams: [2048]
```
tchaton: I would prefer to standardize the API with params:

src/core/data_transform/__init__.py
```diff
@@ -27,17 +27,27 @@ def instantiate_transform(transform_option):
         tr_params = transform_option.params
     except KeyError:
         tr_params = None
+    try:
```
tchaton: Should not be needed here.

nicolas-chaulet: Actually it is, the getattr method fails with a KeyError for OmegaConf dict even if you give a default value

PR #158, opened 2020-02-14
Refactoring scheduler + training

tchaton:
In this PR, schedulers will be refactored to handle more diversity and allow configuration directly from config file

[x] - Allow scheduler creation from config file (lr_scheduler, bn_scheduler)
[x] - Update model checkpoint
[x] - Test model checkpoint
[x] - Allow gradient accumulation
[] - Verify Plateau works with all metric_name


tchaton:
Hey @nicolas-chaulet , could you review this PR too ;)

src/core/schedulers/bn_schedulers.py
```diff
@@ -0,0 +1,95 @@
...
+            * bn_scheduler_params.bn_decay ** (int(it * batch_size / bn_scheduler_params.decay_step)),
+            bn_scheduler_params.bn_clip,
+        )
+    elif bn_scheduler_opt.bn_policy == "exponential_decay":
+        bn_lambda = lambda e: max(
+            eval(bn_scheduler_params.gamma) ** (e / bn_scheduler_params.decay_step),
+            bn_scheduler_params.bn_clip / bn_scheduler_params.bn_momentum,
+        )
+    else:
+        return NotImplementedError("learning rate policy [%s] is not implemented", bn_scheduler_opt.lr_policy)
```
nicolas-chaulet: There are some lr scattered in this file here and there. Please clean.

src/core/schedulers/bn_schedulers.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @property
+    def scheduler_opt(self):
+        return self._scheduler_opt
+
+    @scheduler_opt.setter
+    def scheduler_opt(self, scheduler_opt):
+        self._scheduler_opt = scheduler_opt
+
+    def step(self, epoch=None):
```
nicolas-chaulet: Please add a test

src/core/schedulers/lr_schedulers.py
```diff
@@ -1,46 +1,81 @@
...
 
 
 log = logging.getLogger(__name__)
 
+def repr(self, scheduler_params={}):
+    return "{}({})".format(self.__class__.__name__, scheduler_params)
 
-def build_basic_params():
-    return DictConfig({"base_lr": 0.001})
+class SchedulerWrapper():
```
nicolas-chaulet: Rename to LRScheduler

src/metrics/model_checkpoint.py
```diff
@@ -166,11 +218,12 @@ def save_best_models_under_current_metrics(self, model: BaseModel, metrics_holde
             model {[BaseModel]} -- [Model]
             metrics_holder {[Dict]} -- [Need to contain stage, epoch, current_metrics]
         """
-
         metrics = metrics_holder["current_metrics"]
         stage = metrics_holder["stage"]
         epoch = metrics_holder["epoch"]
 
+        model.set_metrics(metrics, stage, epoch)
```
nicolas-chaulet: Please remove

src/datasets/base_dataset.py
```diff
@@ -55,6 +55,16 @@ def __init__(self, dataset_opt, training_opt):
...
                 setattr(self, new_name, transform)
 
+        self._dataset_state = None
+
+    @property
+    def dataset_state(self):
+        return self._dataset_state
+
+    @dataset_state.setter
+    def dataset_state(self, dataset_state):
```
nicolas-chaulet: I wonder why this is needed. It should not

src/metrics/model_checkpoint.py
```diff
@@ -36,28 +40,34 @@ def get_model_checkpoint(
...
-        self._check_path = "{}.pt".format(check_name)
+        self._check_path = checkpoint_file
         self._initialize_objects()
 
     def _initialize_objects(self):
         self._objects = {}
         self._objects["models"] = {}
+        self._objects["model_state"] = None
         self._objects["stats"] = {"train": [], "test": [], "val": []}
         self._objects["optimizer"] = None
```
nicolas-chaulet: I think `lr_params` is not used anymore. Please remove

src/modules/KPConv/partial_dense.py
```diff
@@ -57,3 +57,109 @@ def __init__(
         # PARAMTERS IMPORTANT FOR SHADOWING
         self.shadow_features_fill = 0.0
         self.shadow_points_fill_ = float(10e6)
+
+
+class KPConvPartialDense(BaseKPConvPartialDense):
```
nicolas-chaulet: Please remove

src/utils/config.py
```diff
@@ -11,6 +11,13 @@
 
 from .enums import ConvolutionFormat
 
+def create_new_omega_conf(opt):
```
nicolas-chaulet: What does this actually do?

src/utils/config.py
```diff
@@ -11,6 +11,13 @@
...
+    empty_opt = OmegaConf.create({})
+    return OmegaConf.merge(opt, empty_opt)
+
+def merge_omega_conf(opt, d):
```
nicolas-chaulet: Same, how is this different from a merge?

test/mockdatasets.py
```diff
@@ -45,6 +46,13 @@ def set_strategies(self, model):
         transform = MultiScaleTransform(strategies)
         self._transform = transform
 
+    @property
+    def dataset_state(self):
```
nicolas-chaulet: Make sure to update if you update the base class

test/test_model_checkpoint.py
```diff
@@ -0,0 +1,139 @@
...
+            shutil.rmtree(ckpt_dir)
+        except:
+            pass
+        if not os.path.exists(ckpt_dir):
+            os.makedirs(ckpt_dir)
+        name = "test"
+        model_checkpoint = ModelCheckpoint(ckpt_dir, name)
+        mock_metrics = {"current_metrics": {"acc": 12}, "stage": "test", "epoch": 10}
+        model_checkpoint.save_best_models_under_current_metrics(model, mock_metrics)
+        model_checkpoint.save_best_models_under_current_metrics(model, mock_metrics)
```
nicolas-chaulet: Please remove the extra one

test/test_model_checkpoint.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Load checkpoint and initialize model
+        model_config = OmegaConf.merge(params, self.data_config)
+        model2 = _find_model_using_name(model_class, "segmentation", model_config, dataset)
+        model_checkpoint = ModelCheckpoint(ckpt_dir, name)
+        model_checkpoint.initialize_model(model2, weight_name="acc")
+        shutil.rmtree(ckpt_dir)
+
+        assert str(model.optimizer.__class__.__name__) == str(model2.optimizer.__class__.__name__)
```
nicolas-chaulet: use self.assert.. instead of the basic assert. It will be a bit clearer when things break

test/test_model_checkpoint.py
```diff
@@ -0,0 +1,139 @@
...
+        model_checkpoint.save_best_models_under_current_metrics(model, mock_metrics)
+        model_checkpoint.save_best_models_under_current_metrics(model, mock_metrics)
+
+        model_checkpoint = ModelCheckpoint(ckpt_dir, name)
+        model = model_checkpoint.create_model_from_checkpoint(dataset)
+        print(model)
+        shutil.rmtree(ckpt_dir)
+
+    """
+    def test_create_model_from_checkpoint_without_dataset(self):
```
nicolas-chaulet: If not used please remove

src/utils/metrics_utils.py
```diff
@@ -0,0 +1,8 @@
+
+def get_from_metric(latest_metrics, query_metric_name, current_stage):
```
nicolas-chaulet: What does this do? Please add a doctstring

train.py
```diff
@@ -153,26 +152,22 @@ def main(cfg):
...
     model_config = OmegaConf.merge(model_config, cfg_training)
     model = instantiate_model(model_class, tested_task, model_config, dataset)
     log.info(model)
 
-    # Optimizer
-    model.set_optimizer(
-        getattr(torch.optim, otimizer_class, None), cfg_training.optimizer.params, cfg_training.learning_rate
-    )
+    # Initialize optimizer, schedulers
+    model.instantiate_optim(cfg)
```
nicolas-chaulet: Rename to instantiate_optimizer

src/models/base_model.py
```diff
@@ -195,24 +294,30 @@ def search_from_key(modules):
...
             return search_with_keys(value, keys[1:], value_out)
+
         try:
             value_out = search_with_keys(opt, keys, value_out)
         except Exception as e:
-            print(e)
+            if msg_err:
+                raise Exception(str(msg_err))
+            else:
+                print(e)
```
nicolas-chaulet: Use `log.exception("Some message")` instead. It will print your message as well as the exception

src/models/base_model.py
```diff
@@ -54,10 +80,25 @@ def lr_params(self):
...
             return None
 
+    @property
+    def schedulers(self):
+        return self._schedulers
+
+    @schedulers.setter
+    def schedulers(self, schedulers):
+        if schedulers:
+            self._schedulers = schedulers
```
nicolas-chaulet: I am wondering if you could have only one way to manage schedulers as opposed to these two ways. The setters and getters are here for that in principle, but at the moment the instantiate_optimi has to know that it should set those two variables when it should not. Maybe a method `set_cheduler(scheduler_name, scheduler)` would help you manage that in a cleaner way.

src/models/base_model.py
```diff
@@ -90,17 +131,52 @@ def get_output(self):
     def forward(self) -> Any:
         """Run forward pass; called by both functions <optimize_parameters> and <test>."""
 
+    def set_selection_stage(self, set_selection_stage):
+        self._set_selection_stage = set_selection_stage
+
+    def set_metrics(self, metrics, stage, epoch):
```
nicolas-chaulet: I think we should remove that

src/models/base_model.py
```diff
@@ -90,17 +131,52 @@ def get_output(self):
     def forward(self) -> Any:
         """Run forward pass; called by both functions <optimize_parameters> and <test>."""
 
+    def set_selection_stage(self, set_selection_stage):
```
nicolas-chaulet: Same I think this should not be there

src/models/base_model.py
```diff
@@ -114,11 +190,32 @@ def get_current_losses(self):
...
+        optimizer_cls = getattr(torch.optim, optmizer_cls_name)
+        optimizer_params = {}
+        if hasattr(optimizer_opt, "params"):
+            optimizer_params = optimizer_opt.params
+        self._optimizer = optimizer_cls(self.parameters(), **optimizer_params)
+        colored_print(COLORS.Green, "Optimizer: {}".format(self._optimizer))
+
+        scheduler_opt = self.get_from_opt(config, ["training", "optim", "scheduler"])
+        if scheduler_opt:
+            self._lr_scheduler = instantiate_scheduler(self._optimizer, scheduler_opt)
```
nicolas-chaulet: See my comment in the schedulers setter.

src/models/base_model.py
```diff
@@ -90,17 +131,52 @@ def get_output(self):
...
         """Calculate losses, gradients, and update network weights; called in every training iteration"""
         self._iterations += batch_size
         self.forward()  # first call forward to calculate intermediate results
-        self._optimizer.zero_grad()  # clear existing gradients
+        make_optimizer_step = self.manage_optimizer_zero_grad()  # Accumulate gradient if option is up
         self.backward()  # calculate gradients
         if self._grad_clip > 0:
             torch.nn.utils.clip_grad_norm_(self.parameters(), self._grad_clip)
-        self._optimizer.step()  # update parameters
+        if make_optimizer_step:
```
nicolas-chaulet: I think this whole logic should be tested. It is really dangerous and could break silently resulting in low performance training which would be really hard to debug.

src/metrics/model_checkpoint.py
```diff
@@ -132,6 +165,25 @@ def __init__(
...
+        raise NotImplementedError
+
+    def create_model_from_checkpoint(self, dataset=None, weight_name=None):
+        if not self._checkpoint.is_empty:
+            model_state = self._checkpoint.get_model_state()
+            model_class = model_state["model_class"]
+            option = OmegaConf.create(model_state["option"])
+            dataset_state = model_state["dataset_state"]
+            task = dataset_state["task"]
+            if not dataset:
```
nicolas-chaulet: How would you be in this position? Is it in preparation for a forward only method? If yes then let's ignore it for now and revisit when we get there. A forward only script would still define a dataset anyway.

src/metrics/model_checkpoint.py
```diff
@@ -132,6 +165,25 @@ def __init__(
...
 
+    def create_mock_dateset(self, dataset_state):
+        raise NotImplementedError
+
+    def create_model_from_checkpoint(self, dataset=None, weight_name=None):
+        if not self._checkpoint.is_empty:
+            model_state = self._checkpoint.get_model_state()
+            model_class = model_state["model_class"]
+            option = OmegaConf.create(model_state["option"])
+            dataset_state = model_state["dataset_state"]
```
nicolas-chaulet: I think I would like to remove this dataset_state for now and think about a better way of handling that. I don't think it is needed for now.

src/metrics/model_checkpoint.py
```diff
@@ -213,4 +266,6 @@ def save_best_models_under_current_metrics(self, model: BaseModel, metrics_holde
                 current_stat[metric_name] = metric_value
                 current_stat["best_{}".format(metric_name)] = metric_value
 
-        self._checkpoint.save_objects(models_to_save, stage, current_stat, model.optimizer, model.lr_params, **kwargs)
+        self._checkpoint.save_objects(
+            models_to_save, model.model_state, stage, current_stat, model.optimizer, model.schedulers, **kwargs
```
nicolas-chaulet: It is a bit weird to have this second model_state (models to save are model states as well). Any chance this new property could come from the `state_dict` method? Like that you have one function that defines the state of the model and that's it.

conf/models/segmentation/kpconv.yaml
```diff
@@ -22,29 +22,6 @@ models:
             nn: [64, 64, 64, 64, 64]
             dropout: 0.5
 
-    PDSimpleKPConv:
```
nicolas-chaulet: You need that one for the tests

conf/training/default.yaml
```diff
@@ -1,22 +1,27 @@
...
-            lr: 0.001
-    learning_rate:
-        scheduler_type: "step_decay"
-        base_lr: ${optimizer.params.lr}
-        lr_decay: 0.5
-        decay_step: 200000
-        lr_clip: 1e-5
+    optim:
+        base_lr: 0.001
+        accumulated_gradient: 10 # Accumulate gradient accumulated_gradient * batch_size
```
nicolas-chaulet: Should this not be -1 by default?

src/__init__.py
```diff
@@ -47,11 +51,17 @@ def instantiate_model(model_class, task, option, dataset: BaseDataset) -> BaseMo
...
         raise NotImplementedError(
             "In %s.py, there should be a subclass of BaseDataset with class name that matches %s in lowercase."
             % (model_module, class_name)
         )
-    return model(option, "dummy", dataset, modellib)
+
+    option_container = OmegaConf.to_container(option)
+    model = model_cls(option, "dummy", dataset, modellib)
+
+    model_state = {"model_class": model_class, "option": option_container, "model_module": model_module, "task": task}
```
nicolas-chaulet: As per previous comment did you consider avoiding this extra state? In addition to the state_dict provided by pytorch? In my opinion there should only be one state to store all information (the module state_dict seems like a good candidate)

src/metrics/model_checkpoint.py
nicolas-chaulet: Don't you think this can be confusing?

src/models/base_model.py
```diff
@@ -43,7 +47,27 @@ def __init__(self, opt):
...
+        self._accumulated_gradient_step = self.get_from_opt(opt, ["optim", "accumulated_gradient"])
+        if self._accumulated_gradient_step:
+            if self._accumulated_gradient_step > 1:
+                self._accumulated_gradient_count = 0
+                colored_print(COLORS.Green, "Accumulated option activated {}".format(self._accumulated_gradient_step))
+            else:
+                raise Exception("accumulated_gradient should be greater than 1")
+
+    @property
+    def model_state(self):
```
nicolas-chaulet: See my previous comment. I think this one should be merged with pytorch state_dict variable.

test/test_models.py
```diff
@@ -71,28 +71,8 @@ def test_kpconv(self):
         model.forward()
         model.backward()
 
-    def test_kpconvpretransform(self):
```
nicolas-chaulet: Could you please add this test back?

src/utils/config.py
```diff
@@ -8,9 +8,41 @@
...
+    selection_stage = getattr(cfg, "selection_stage", None)
+    if not selection_stage:
+        selection_stage = "val" if has_val_loader else "test"
+    else:
+        if not has_val_loader and selection_stage == "val": 
+            raise Exception("Selection stage should be: test")
+    return selection_stage
+
+def merge_omega_conf(opt: DictConfig, d: dict):
+    """This function allows to merge a OmegaConf DictConfig with a python dictionnary"""
```
nicolas-chaulet: Nice thanks

src/models/base_model.py
```diff
@@ -43,7 +47,34 @@ def __init__(self, opt):
...
```
nicolas-chaulet: Please remove those ones

PR #161, opened 2020-02-17
Deformable fix

nicolas-chaulet:
Cleaning up KPConv


src/models/base_model.py
```diff
@@ -134,29 +138,29 @@ def get_named_internal_losses(self):
...
         """
         losses_global = defaultdict(list)
+
         def search_from_key(modules, losses_global):
             for _, module in modules.items():
                 if isinstance(module, BaseInternalLossModule):
                     losses = module.get_internal_losses()
                     for loss_name, loss_value in losses.items():
-                        losses_global[loss_name].append(loss_value)
+                        if isinstance(loss_value, torch.Tensor):
```
tchaton: use torch.is_tensor there

src/modules/KPConv/kernels.py
```diff
@@ -1,144 +1,18 @@
...
-        return aggr_out
-
-    def __repr__(self):
-        # PointKernel parameters
-        return "PointKernel({}, {}, {}, {}, {})".format(
-            self.in_features, self.out_features, self.num_points, self.radius, self.KP_influence,
-        )
-
-
 class LossKPConvEnum(Enum):
```
tchaton: Do we still need this ?

src/modules/KPConv/kernels.py
```diff
@@ -1,144 +1,18 @@
...
     PERMISSIVE = 0
     FITTING = 1
     REPULSION = 2
 
+
 def fitting_loss(sq_distance, radius):
```
tchaton: Docstring on losses would be nice, and maybe move them to the losses file.

PR #166, opened 2020-02-19
Checkpoint

nicolas-chaulet:
Second shot


eval.py
```diff
@@ -63,57 +61,31 @@ def run(cfg, model, dataset: BaseDataset, device, tracker: BaseTracker, checkpoi
...
-    torch.backends.cudnn.enabled = cfg_eval.enable_cudnn
+    torch.backends.cudnn.enabled = cfg.enable_cudnn
 
-    # Find and create associated dataset
-    dataset_config = cfg.data
-    dataset_class = getattr(dataset_config, "class")
-    dataset_config.dataroot = hydra.utils.to_absolute_path(dataset_config.dataroot)
-    dataset = instantiate_dataset(dataset_class, tested_task, dataset_config, cfg_eval)
+    # Checkpoint
+    checkpoint = ModelCheckpoint(cfg.checkpoint_dir, cfg.model_name, False, cfg.weight_name)
```
tchaton: If cfg.model_name is wrong, print the correct name in the error msg

src/datasets/base_dataset.py
```diff
@@ -58,54 +59,56 @@ def __init__(self, dataset_opt, training_opt):
...
+            conv_type.lower() == ConvolutionFormat.PARTIAL_DENSE.value[-1].lower()
+            or conv_type.lower() == ConvolutionFormat.MESSAGE_PASSING.value[-1].lower()
         ):
             return lambda datalist: torch_geometric.data.batch.Batch.from_data_list(datalist)
-        elif conv_type == ConvolutionFormat.DENSE.value[-1].lower():
+        elif conv_type.lower() == ConvolutionFormat.DENSE.value[-1].lower():
             return lambda datalist: SimpleBatch.from_data_list(datalist)
 
-    def _create_dataloaders(self, train_dataset, test_dataset, val_dataset=None):
+    def create_dataloaders(
```
tchaton: I think this interface is a bit less intuitive than the previous one. Anyway to make it leaner ?

train.py
```diff
@@ -153,80 +152,49 @@ def main(cfg):
...
-    model.instantiate_optimizers(cfg)
+    # Create model and datasets
+    if not checkpoint.is_empty and cfg.training.resume:
+        dataset = instantiate_dataset(checkpoint.data_config)
+        model = checkpoint.create_model(dataset, weight_name=cfg.training.weight_name)
+    else:
+        dataset = instantiate_dataset(cfg.data)
+        model = instantiate_model(cfg, dataset)
+        model.instantiate_optimizers(cfg)
+
```
tchaton: Would be nice to log both model and dataset there.

eval.py
nicolas-chaulet: It would log something like that now:
```
ValueError: The provided path /workspaces/deeppointcloud-benchmark/outputs/2020-02-19/15-54-37 didn't contain the checkpoint_file pointnet2__.pt
Did you mean pointnet2.pt
```
I have also forced the checkpoint to not be empty in case we are in resume mode.

src/datasets/base_dataset.py
nicolas-chaulet: Do you have anything in mind? It needs all the inputs and not more in principle...

src/datasets/base_dataset.py
```diff
@@ -218,3 +213,11 @@ def set_strategies(self, model):
     @abstractmethod
     def get_tracker(model, dataset, wandb_log: bool, tensorboard_log: bool):
         pass
+
+    def __repr__(self):
+        message = "Dataset: %s \n" % self.__class__
```
tchaton: Please, use self.__class__.__name__ instead
Also, add num_elements per classes et batch size

PR #167, opened 2020-02-19
Panel

tchaton:


pyproject.toml
```diff
@@ -20,6 +20,10 @@ torch-points = {git = "https://github.com/nicolas-chaulet/torch-points.git", rev
 torch-scatter = "^1.4.0"
 torch-sparse = "==0.4.3"
 torch-cluster = "^1.4.5"
+jupyterlab = "^1.2.6"
```
nicolas-chaulet: Ad I said, add those as dev dependencies (poetry add --dev)

src/visualization/experiment_manager.py
```diff
@@ -0,0 +1,150 @@
+import os
+from glob import glob
+from collections import defaultdict
+import torch
+
+from src.utils.colors import COLORS
+
```
nicolas-chaulet: Nice one, could you rework the script that uses this code then to avoid duplicates?

src/visualization/experiment_manager.py
```diff
@@ -0,0 +1,150 @@
...
+
+def colored_print(color, msg):
```
nicolas-chaulet: Would you mind using the utils? Like that it will be plugged in the logger by default.

PR #176, opened 2020-02-25
add support for multi test

tchaton:


eval.py
```diff
@@ -90,6 +100,7 @@ def main(cfg):
     )
     log.info(dataset)
 
+    model.eval()
```
nicolas-chaulet: Good spot

src/datasets/base_dataset.py
```diff
@@ -88,9 +88,15 @@ def create_dataloaders(
             sampler=self.train_sampler,
         )
 
-        self._test_loader = dataloader(
-            self.test_dataset, batch_size=batch_size, shuffle=False, num_workers=num_workers, sampler=self.test_sampler,
-        )
+        if isinstance(self.test_dataset, list):
+            log.warning('Please, make sure you have set dataset_name and set selection_stage to one of them to properly track the best model')
```
nicolas-chaulet: Rewrite a bit:
'You are using multiple test sets, please, make sure you have set dataset_name and set selection_stage to one of them to properly track the best model. Choose amongst ...'

src/datasets/base_dataset.py
```diff
@@ -118,6 +124,13 @@ def val_dataloader(self):
     def test_dataloader(self):
         return self._test_loader
 
+    @property
+    def num_test_datasets(self):
+        if isinstance(self._test_loader, list):
+            return self._test_loader
```
nicolas-chaulet: `return len()`

eval.py
```diff
@@ -41,20 +40,31 @@ def eval_epoch(model: BaseModel, dataset, device, tracker: BaseTracker, checkpoi
...
-        for data in tq_test_loader:
-            data = data.to(device)
-            with torch.no_grad():
-                model.set_input(data)
-                model.forward()
 
-            tracker.track(model)
-            tq_test_loader.set_postfix(**tracker.get_metrics(), color=COLORS.TEST_COLOR)
+    loaders = dataset.test_dataloader()
+    if not isinstance(loaders, list):
```
nicolas-chaulet: I think this should be handled by the dataset, it would always build a list of datasets no matter what

eval.py
```diff
@@ -41,20 +40,31 @@ def eval_epoch(model: BaseModel, dataset, device, tracker: BaseTracker, checkpoi
...
+        loaders = [loaders]
 
-    tracker.print_summary()
+    num_loaders = len(loaders)
```
nicolas-chaulet: Use the model's property to get the number of data loaders

src/datasets/base_dataset.py
```diff
@@ -88,9 +88,15 @@ def create_dataloaders(
...
-        self._test_loader = dataloader(
-            self.test_dataset, batch_size=batch_size, shuffle=False, num_workers=num_workers, sampler=self.test_sampler,
-        )
+        if isinstance(self.test_dataset, list):
+            log.warning('Please, make sure you have set dataset_name and set selection_stage to one of them to properly track the best model')
+            self._test_loader = [dataloader(dataset, batch_size=batch_size, shuffle=False, 
+                                num_workers=num_workers, sampler=self.test_sampler,
+                                ) for dataset in self.test_dataset]
+        else:
+            self._test_loader = dataloader(
```
nicolas-chaulet: Always build a list

src/datasets/base_dataset.py
```diff
@@ -183,25 +196,42 @@ def batch_size(self):
...
     @property
     def num_batches(self):
-        return {
+        out = {
             "train": len(self._train_loader),
-            "test": len(self._test_loader),
             "val": len(self._val_loader) if self.has_val_loader else 0,
         }
+        stage_name = "test"
+        if isinstance(self._test_loader, list):
```
nicolas-chaulet: I wouldn't have an if here. Treat the single and multiple case the same way.

src/datasets/base_dataset.py
```diff
@@ -183,25 +196,42 @@ def batch_size(self):
...
-                    setattr(attr.dataset, "transform", transform)
-                else:
-                    if isinstance(current_transform, Compose):  # The transform contains several transformations
-                        current_transform.transforms += [transform]
-                    else:
-                        setattr(
-                            attr.dataset, "transform", Compose([current_transform, transform]),
-                        )
+        for attr_name, attr in self.__dict__.items():
+            if isinstance(attr, list):
```
nicolas-chaulet: Same

src/datasets/base_dataset.py
```diff
@@ -222,7 +252,9 @@ def __repr__(self):
         for attr in self.__dict__:
             if attr.endswith("_dataset"):
                 dataset = getattr(self, attr)
-                if dataset:
+                if isinstance(dataset, list):
+                    size = ",".join([str(len(d)) for d in dataset])
```
nicolas-chaulet: Same

src/utils/config.py
```diff
@@ -53,20 +53,19 @@ def launch_wandb(cfg, launch: bool):
...
        is going to be on the validation or test dataset
        keys: ["test", "val"]
     """
     selection_stage = getattr(cfg, "selection_stage", None)
     if not selection_stage:
-        selection_stage = "val" if has_val_loader else "test"
-    else:
-        if not has_val_loader and selection_stage == "val":
-            raise Exception("Selection stage should be: test")
+        if num_test_datasets == 1:
```
nicolas-chaulet: Same

src/visualization/visualizer.py
```diff
@@ -80,11 +83,11 @@ def get_indices(self, stage):
                 if self._deterministic:
                     if stage not in self._indices:
                         if self._num_samples_per_epoch > total_items:
-                            raise Exception("Number of samples to save if higher than the number of available elements")
+                            log.warn("Number of samples to save is higher than the number of available elements")
```
nicolas-chaulet: In case _num_samples_per_epoch. total_items then I suggest that we log the warning but also set _num_samples_per_epoch to total_items. Like that it will generate visuals no matter what

test/test_dataloader_list.py
```diff
@@ -0,0 +1,75 @@
...
+        mock_base_dataset = MockBaseDataset(dataset_opt)
+
+        model_config = MockModelConfig()
+        setattr(model_config, "conv_type", "dense")
+        model = MockModel(model_config)
+
+        mock_base_dataset.create_dataloaders(model, 2, True, 0, False)
+        datasets = mock_base_dataset.test_dataloader()
+
+        self.assertEqual(type(datasets), list)
```
nicolas-chaulet: Test the length rather than the type

test/test_dataloader_list.py
```diff
@@ -0,0 +1,75 @@
...
+        mock_base_dataset = MockBaseDataset(dataset_opt)
+
+        model_config = MockModelConfig()
+        setattr(model_config, "conv_type", "dense")
+        model = MockModel(model_config)
+
+        mock_base_dataset.create_dataloaders(model, 2, True, 0, False)
+        datasets = mock_base_dataset.test_dataloader()
+
+        self.assertIsNot(type(datasets), list)
```
nicolas-chaulet: They will all be lists, check the length (as well as the property that should return the number of test datasets)

train.py
```diff
@@ -65,6 +65,8 @@ def train_epoch(
 
             iter_data_time = time.time()
 
+            break
```
nicolas-chaulet: Sure about that

train.py
```diff
@@ -96,6 +98,8 @@ def eval_epoch(
             if visualizer.is_active:
                 visualizer.save_visuals(model.get_current_visuals())
 
+            break
```
nicolas-chaulet: Sure about that?

train.py
```diff
@@ -111,25 +115,39 @@ def test_epoch(
...
-        for data in tq_test_loader:
-            data = data.to(device)
-            with torch.no_grad():
-                model.set_input(data)
-                model.forward()
 
-            tracker.track(model)
-            tq_test_loader.set_postfix(**tracker.get_metrics(), color=COLORS.TEST_COLOR)
+    loaders = dataset.test_dataloader()
+    if not isinstance(loaders, list):
```
nicolas-chaulet: Same change to deal with lists no matter what

src/utils/config.py
```diff
@@ -53,20 +53,19 @@ def launch_wandb(cfg, launch: bool):
...
+            selection_stage = "val" if has_val_loader else "test"
+        else:
+            selection_stage = "val" if has_val_loader else "test:0"
```
nicolas-chaulet: Add a static method to the base_dataset that builds this string. Could be something like that:
```
def get_dataset_name(stage, index=0):
    return "stage:index"
```
It will make sure that this "convention" is respected throughout the code base.

src/datasets/base_dataset.py
```diff
@@ -118,6 +124,13 @@ def val_dataloader(self):
     def test_dataloader(self):
         return self._test_loader
```
nicolas-chaulet: Actually, better idea, deal with the fact that it may not always be a list here. if self._test_loader is not a list this property will return [self._test_loader]. Like that all existing datasets will still be working and the interface is clear: test_dataloader is a list.

conf/config.yaml
```diff
@@ -19,5 +19,5 @@ defaults:
 
 model_name: ??? # Name of the specific model to load
 
-selection_stage: "test" # Which dataset to use to select best model ["val", "test"]
+selection_stage: ""
```
nicolas-chaulet: Good spot!

conf/data/segmentation/shapenet.yaml
```diff
@@ -6,6 +6,7 @@ data:
     first_subsampling: 0.02
     train_size: 14016 # Number of shapes in the whole training set
     use_category: True
+    category: 'Airplane'
```
nicolas-chaulet: Please remove

eval.py
```diff
@@ -37,24 +36,29 @@ def eval_epoch(model: BaseModel, dataset, device, tracker: BaseTracker, checkpoi
             tracker.track(model)
             tq_val_loader.set_postfix(**tracker.get_metrics(), color=COLORS.VAL_COLOR)
 
+            break
```
nicolas-chaulet: Please remove

find_neighbour_dist.py
```diff
@@ -59,8 +59,9 @@ def run(cfg, model: BaseModel, dataset: BaseDataset, device, measurement_name: s
         run_epoch(model, dataset.val_dataloader(), device, num_batches)
         measurements["val"] = extract_histogram(model.get_spatial_ops(), normalize=False)
 
-    run_epoch(model, dataset.test_dataloader(), device, num_batches)
-    measurements["test"] = extract_histogram(model.get_spatial_ops(), normalize=False)
+    for loader_idx, loader in enumerate(dataset.test_dataloaders()):
+        run_epoch(model, dataset.test_dataloaders(), device, num_batches)
+        measurements["test:{}".format(loader_idx)] = extract_histogram(model.get_spatial_ops(), normalize=False)
```
nicolas-chaulet: Use the dataset get stage name instead of that.

src/datasets/base_dataset.py
```diff
@@ -181,31 +199,62 @@ def feature_dimension(self):
...
+        if hasattr(loader.dataset, "dataset_name"):
+            return loader.dataset.dataset_name
+        else:
+            if self.num_test_datasets > 1:
+                return "test_{}".format(index)
+            else:
+                return"test"
+
+    @property
+    def determine_stage(self):
```
nicolas-chaulet: It does not feel right to have this as a property. See my comment a bit below, it would make sense to move the method that is in the config.py here and make it a static method. That will be actually nice.

src/utils/config.py
```diff
@@ -53,20 +53,17 @@ def launch_wandb(cfg, launch: bool):
...
        keys: ["test", "val"]
     """
     selection_stage = getattr(cfg, "selection_stage", None)
-    if not selection_stage:
-        selection_stage = "val" if has_val_loader else "test"
-    else:
-        if not has_val_loader and selection_stage == "val":
-            raise Exception("Selection stage should be: test")
+    if not selection_stage or selection_stage != "":
+        selection_stage = dataset.determine_stage
```
nicolas-chaulet: I think this can be reviewed a bit. Something like if dataset.has_val_loader then... else selection_stage=dataset.get_test_name(0)
By doing so you can remove the determine_stage from the base_dataset (which is a bit strange as is)

src/utils/config.py
```diff
@@ -53,20 +53,17 @@ def launch_wandb(cfg, launch: bool):
         wandb.save(os.path.join(os.getcwd(), ".hydra/overrides.yaml"))
 
 
-def determine_stage(cfg, has_val_loader):
+def determine_stage(cfg, dataset, has_val_loader):
```
nicolas-chaulet: has_val_loader is not used anymore

nicolas-chaulet: Also please rename method to `resolve_save_stage`

nicolas-chaulet: Last suggestion, make it a static method of the dataset. That's really where it belongs

src/datasets/base_dataset.py
```diff
@@ -115,16 +122,32 @@ def has_val_loader(self):
...
+
 
     @property
     def has_fixed_points_transform(self):
         """
         This property checks if the dataset contains T.FixedPoints transform, meaning the number of points is fixed
         """
         transform_train = self._train_loader.dataset.transform
-        transform_test = self._test_loader.dataset.transform
+        transform_test = self._test_loaders.dataset.transform
```
nicolas-chaulet: [0]

nicolas-chaulet: self.test_transform

src/datasets/base_dataset.py
```diff
@@ -214,6 +256,28 @@ def set_strategies(self, model):
     def get_tracker(model, dataset, wandb_log: bool, tensorboard_log: bool):
         pass
 
+    def resolve_saving_stage(self, cfg):
```
nicolas-chaulet: `selection_stage`

PR #179, opened 2020-02-28
Updating torch points

nicolas-chaulet:
This PR updates torch_points with a full CPU implementation of all kernels. Ball query now return the distances as well in all cases (CPU and GPU)

### DIFF
--- a/src/core/losses/dirichlet_loss.py
+++ b/src/core/losses/dirichlet_loss.py
@@ -49,7 +49,7 @@ def _dirichlet_dense(r, pos, f, aggr):
 
 
 def _variance_estimator_dense(r, pos, f):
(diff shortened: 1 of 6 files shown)


src/core/losses/dirichlet_loss.py
```diff
@@ -49,7 +49,7 @@ def _dirichlet_dense(r, pos, f, aggr):
 
 
 def _variance_estimator_dense(r, pos, f):
-    nei_idx = tp.ball_query(r, _MAX_NEIGHBOURS, pos, pos).reshape(pos.shape[0], -1).long()  # [B,N * nei]
+    nei_idx = tp.ball_query(r, _MAX_NEIGHBOURS, pos, pos, sort=True)[0].reshape(pos.shape[0], -1).long()  # [B,N * nei]
```
tchaton: Why do you use sort there ?

PR #180, opened 2020-02-29
To train on 3DMatch with pointnet++

humanpose1:
I've corrected some mistakes and I have an example of training using a pointnet++ like model. There are still some problems of convergence (metric learning on patches is hard...) but now we can launch the training.
```
python train.py task=registration model_type=pointnet2 model_name=minipointnet2 dataset=patch3dmatch training=default_reg lr_scheduler=multi_step_reg
```


conf/lr_scheduler/multi_step_reg.yaml
```diff
@@ -0,0 +1,5 @@
+lr_scheduler:
+  class: MultiStepLR
+  params:
+    milestones: [3000, 6000, 9000, 15000, 21000, 30000]
```
tchaton: The BaseModel is providing epoch and not steps anymore. Those won t be reached

src/datasets/registration/pair.py
```diff
@@ -0,0 +1,21 @@
+import torch
+from torch_geometric.data import Data
+from torch_geometric.data import Batch
+
+
+def make_pair(data_source: Data, data_target: Data):
```
tchaton: Maybe this one should be moved to some datasets utile along with others batch aggregations

src/core/data_transform/features.py
```diff
@@ -0,0 +1,71 @@
+import torch
+
+
+class OneFeature(object):
```
tchaton: Why do u need this ? If it is KPConv, I will add this within a PR on Monday ?

src/core/data_transform/features.py
```diff
@@ -0,0 +1,71 @@
...
+        assert data.pos is not None
+        xyz = data.pos[:, self.axis]
+        if data.x is None:
+            data.x = xyz
+        else:
+            data.x = torch.cat([data.x, xyz], -1)
+        return data
+
+
+class RGBFeature(object):
```
tchaton: Add normalize option to put rgb within 0, 1

tchaton:
@humanpose1, awesome PR ! I will give it a try :)

src/core/data_transform/features.py
```diff
@@ -0,0 +1,71 @@
...
+        return data
+
+
+class NormalFeature(object):
+    """
+    add normal as feature. if it doesn't exist, compute normals
+    using PCA
+    """
+    def __call__(self, data):
+        if data.norm is None:
```
nicolas-chaulet: rename to data.normal

humanpose1: As you want but, I use norm because in torch_geometric they use also that [name.](https://github.com/rusty1s/pytorch_geometric/blob/18da46c2a28faad83aca9fb7602500c69fc2d4b0/torch_geometric/data/data.py#L47)

nicolas-chaulet: Alright, leave it as is then! I did not know that

Issue #181, opened 2020-02-29
how to test custom dataset （pointcloud） using kpconv model.


Yuwenger:
@nicolas-chaulet 
 sorry to bother you, I defined my own dataset and trained. but how can i test the result of model.(I mean a full pointcloud pointwist prediction, in eval.py or in viz folder the data is part of a single item(mostly a slice of pointcloud within  a defined radius)).

tchaton:
Hey @Yuwenger, excellent question! We didn t implement a prediction mapping for slices to full pointcloud. We will have to do it, but it isn t our priority right now. If you wish, you could add the functionality and make a PR. We could try to help you if needed. But it is a bit tricky to do correctly depending on pointcloud size (if it doesn t fit in memory and you have sampling transform, using Knn to interpolate predictions).

Yuwenger:
hey @tchaton ，thanks for your reply ,I will deep in the code and have a try by myself. And look forward to your release.

tchaton:
Hey @Yuwenger, we hava a Slack where we discuss about the project with several researchers.
Do you wanna join ? If yes, I will need your email adress ;)

Yuwenger:
Hey @tchaton ,I am glad to join it .  My email adress is: gu_yu_feng@sina.com

Yuwenger:
Hey @tchaton  ,I am glad to join it . My email adress will send to your email or you can get it from my github homepage profile.

tchaton:
@Yuwenger Did you manage to run KPConv on your custom dataset ?

tchaton:
https://github.com/nicolas-chaulet/torch-points3d/blob/master/notebooks/PartSegmentationKPConv.ipynb

PR #184, opened 2020-03-04
Forwardshapenet

nicolas-chaulet:
Working forward shapenet with dashboard for Viz 
This may be a way to go for #181


forward_scripts/conf/partseg.yaml
```diff
@@ -0,0 +1,11 @@
+num_workers: 2
```
tchaton: rename shapenet_forward

PR #187, opened 2020-03-05
Kpconv fix

nicolas-chaulet:
Fix a problem with grid size definitions in KPConv
(Also adds caching mechanism to the ci)


src/modules/KPConv/blocks.py
```diff
@@ -238,15 +253,22 @@ def __init__(
...
+                block_kwargs[key] = arg[i] if is_list(arg) else arg
+
+            # Building the block
             kpcls = getattr(sys.modules[__name__], class_name)
             block = kpcls(
                 down_conv_nn=down_conv_nn[i],
                 grid_size=grid_size[i],
-                is_strided=is_strided[i],
+                prev_grid_size=prev_grid_size[i],
                 has_bottleneck=has_bottleneck[i],
```
tchaton: Would be great to add this is_list to all of them.

PR #201, opened 2020-03-25
Testing all models for forward

nicolas-chaulet:
This PR is mainly about running a foward and backward on all of our models.

@humanpose1 I have modified your registration models slightly, because they were failing to forward in the tests (probably invalid kind of data). Please double check that I haven't broken anything!


conf/models/registration/pointnet2.yaml
```diff
@@ -8,7 +8,7 @@ models:
       radii: [[0.2], [0.4]]
       nsamples: [[64], [64]]
       down_conv_nn:
-        [[[FEAT, 32, 32, 64]], [[64+3, 64, 128, 128]]]
+        [[[FEAT +3, 32, 32, 64]], [[64+3, 64, 128, 128]]]
```
humanpose1: The problem here is that the feature here is only XYZ. I got an error when data.x was empty...

nicolas-chaulet: It shouldn't anymore, let me know if you still see it

src/models/registration/pointnet2.py
```diff
@@ -91,4 +91,5 @@ def backward(self):
         """Calculate losses, gradients, and update network weights; called in every training iteration"""
         # caculate the intermediate results if necessary; here self.output has been computed during function <forward>
         # calculate loss given the input and intermediate results
-        self.loss.backward()  # calculate gradients of network G w.r.t. loss_G
+        if hasattr(self, "loss") and self.loss:
```
tchaton: Are you sure you can do that ? Won't it trigger an error if you don't have loss in self.

conf/models/registration/pointnet2.yaml
humanpose1: I know why you had this problem. I add XYZ in the transforms whereas you add in the network(I put use_xyz=False)

test/mockdatasets.py
```diff
@@ -18,10 +18,10 @@ def get(self, dataset_name, default):
 
 
 class MockDataset(torch.utils.data.Dataset):
-    def __init__(self, feature_size=0):
+    def __init__(self, feature_size=0, transform=None, num_points=100):
```
tchaton: I would prefer to have input_nc, output_nc there. Because by default, this mock dataset works only for segmentation. Would be great to have as generic as possible.

test/test_base_dataset.py
```diff
@@ -1,149 +0,0 @@
...
-        self.assertEqual(dataset.test_dataset, None)
-        self.assertEqual(dataset.val_dataset, None)
-
-    def test_simple_datasets(self):
-
-        opt = Options()
-        opt.dataset_name = os.path.join(os.getcwd(), "test")
-        opt.dataroot = os.path.join(os.getcwd(), "test")
-        
-        class SimpleDataset(BaseDataset):
```
tchaton: I liked using directly BaseDataset for Mocking the dataset. I think it would bring more robustness within the code.

test/test_models.py
```diff
@@ -41,26 +42,46 @@ def test_createall(self):
...
-                    models_config.update("model_name", model_name)
-                    instantiate_model(models_config, MockDatasetGeometric(6))
+                with self.subTest(model_name):
+                    if model_name not in ["MyTemplateModel", "MinkUNet_WIP"]:
+                        models_config.update("model_name", model_name)
+                        instantiate_model(models_config, MockDatasetGeometric(6))
+
+    def test_runall(self):
+        def is_known_to_fail(model_name):
+            forward_failing = ["MyTemplateModel", "MinkUNet_WIP", "pointcnn", "RSConv_4LD", "RSConv_2LD", "randlanet"]
```
tchaton: MyTemplateModel doesn't t exist anymore I think.

test/test_base_dataset.py
nicolas-chaulet: I haven't deleted your code, just moved it to a different place

test/test_models.py
nicolas-chaulet: ok cool

src/models/registration/pointnet2.py
humanpose1: You can see[here](https://github.com/nicolas-chaulet/deeppointcloud-benchmarks/blob/74784e2b37f2c7c24bc819eee9deef00a4152dda/src/models/base_architectures/backbone.py#L101), where we add the loss. If in the config, no loss has been provided, it will be None. You're right. We should throw an exception if the loss is None (The miner can be None)

test/mockdatasets.py
nicolas-chaulet: Agreed, let's do that when we need to use it for another task

src/models/registration/pointnet2.py
nicolas-chaulet: Nope, it is tested :)

nicolas-chaulet: The only way the loss does not exists is if the forward functions stops earlier (there is a return half way through). I'll leave the hasattr but remove the second check so that it throws an exception

PR #202, opened 2020-03-25
New model

tchaton:


src/modules/MinkowskiEngine/modules.py
```diff
@@ -4,6 +4,55 @@
 from src.utils.config import is_list
 
 
+class SparseResBlock(nn.Module):
```
nicolas-chaulet: Could you please combine that with the other blocks? Or explain why it is different from the others? Also, please review the naming of variables, it is a bit criptic at the moment

src/modules/MinkowskiEngine/unet.py
```diff
@@ -0,0 +1,129 @@
+import torch.nn as nn
+import MinkowskiEngine as ME
+
+from .modules import SparseResBlock
+from src.utils.config import is_list
+
+
+class Sequential(nn.Sequential):
```
nicolas-chaulet: This one should go in core, not here

src/models/segmentation/minkowski.py
```diff
@@ -13,13 +13,44 @@
...
+    def forward(self):
+
+        self.output = self.model(self.input).feats
+        self.loss_seg = F.cross_entropy(self.output, self.labels)
+
+    def backward(self):
+        self.loss_seg.backward()
 
+
+class Minkowski_Model(BaseModel):
```
nicolas-chaulet: Could you please quickly document the difference between each one of these models? Why do we have 3 of them?

src/models/segmentation/minkowski.py
```diff
@@ -13,13 +13,44 @@
...
+        
+        self.model = initialize_baseline_minkowski_unet(
             option.model_name, dataset.feature_dimension, dataset.num_classes, option.D
         )
         self.loss_names = ["loss_seg"]
 
     def set_input(self, data, device):
+        self.batch_idx = data.batch.squeeze()
+        coords = torch.cat([data.batch.unsqueeze(-1), data.pos], -1).int()
+        self.input = ME.SparseTensor(data.x, coords=coords).to(device)
```
nicolas-chaulet: As per current implementation of the full_res functionality, input should contain the raw input data. We can change taht in the future but for now it is kind of a reserved variable. Please use sparse_input for this sparse tensor also checkout fle for an example, I have made the change for the regression model.

PR #206, opened 2020-03-27
'add downloading of scannet'

tchaton:
This PR adds ScanNet dataset with with automatic download + bboxes (optional) + instance segmentation (optional) + scannet_tracker.


src/datasets/segmentation/scannet.py
```diff
@@ -0,0 +1,574 @@
...
+        36: (227., 119., 194.),
+        37: (213., 92., 176.),
+        38: (94., 106., 211.),
+        39: (82., 84., 163.),
+        40: (100., 85., 144.),
+    }
+
+    SPLIT = ["train", "val"]
+
+    DONOTCARE_CLASS_IDS = np.array([])
```
nicolas-chaulet: Review class attributes

src/metrics/scannet_tracker.py
```diff
@@ -0,0 +1,73 @@
+from typing import Dict
```
nicolas-chaulet: Not sure we need that at all if we implement a label "ignore"
-> use -100 https://pytorch.org/docs/stable/nn.html#nllloss

PR #207, opened 2020-03-28
Descriptors/debug patch

humanpose1:
Sorry for this late and big PR...

It is not perfect and there are still room for improvements. But I've started to train and the results are not so bad (even if it's still far from the SOTA... I'm investigating).
I've modified some base_classes (for example I add the possibility to add filter which is important in my case).


src/core/data_transform/features.py
```diff
@@ -166,67 +166,63 @@ def __repr__(self):
...
+        pos_centered = data.pos - data.pos.mean(axis=0)
+        cov_matrix = pos_centered.T.mm(pos_centered) / len(pos_centered)
+        eig, v = torch.symeig(cov_matrix, eigenvectors=True)
+        data.eigenvalues = eig
+        data.eigenvectors = v
         return data
 
 
-class NormalFeature(object):
+class PlanarityFilter(object):
```
nicolas-chaulet, line 214: Looks like it's a duplicate of the filter.py PlanarityFilter

src/datasets/base_dataset.py
```diff
@@ -113,6 +116,21 @@ def set_transform(obj, dataset_opt):
         inference_transform = BaseDataset.add_transform(obj.test_transform, out=inference_transform)
         obj.inference_transform = Compose(inference_transform) if len(inference_transform) > 0 else None
 
+    @staticmethod
+    def set_filter(obj, dataset_opt):
```
nicolas-chaulet, line 120: Please make this method private since it is never called outisde this class

src/datasets/registration/base_siamese_dataset.py
```diff
@@ -0,0 +1,54 @@
+import os
+from abc import ABC, abstractmethod
+import logging
+from functools import partial
+
+import torch
+import torch_geometric
+from torch_geometric.transforms import Compose, FixedPoints
+
+from src.core.data_transform import instantiate_transforms, MultiScaleTransform
```
nicolas-chaulet, line 10: Please clean up the import, import only what you need

src/datasets/registration/pair.py
```diff
@@ -1,21 +1,208 @@
 import torch
+from typing import List, Optional
 from torch_geometric.data import Data
 from torch_geometric.data import Batch
+from src.datasets.multiscale_data import MultiScaleBatch
+import re
 
+class Pair(Data):
```
nicolas-chaulet, line 8: It would be good to test all this

src/models/base_model.py
```diff
@@ -270,6 +271,19 @@ def get_internal_loss(self):
         else:
             return loss / c
 
+    def get_loss_and_miner(self, opt_loss, opt_miner):
```
nicolas-chaulet, line 274: Pleaser make it a static method and document what the config should be

test_registration_scripts/conf/config.yaml
```diff
@@ -0,0 +1,5 @@
+
```
nicolas-chaulet, line 1: Could you move all this in a subfolder within the existing `script` folder?

src/datasets/registration/pair.py
humanpose1: You're right. I'm still not used to write test... Don't hesitate to tell me if there are other things to test ;-)

nicolas-chaulet: I would say that more or less anything that does not need the actual data and can be mocked should be tested to some extent. For example we run a forward and backward pass on all the models with some dummy data. Things that are used in multiple places are more critical, for example, all the transforms are tested as well, most of the basedataset and basemodel. Then we make a judgment call, getting 100% tested would be too troublesome to maintain and probably does not make too much sense.

src/core/data_transform/features.py
```diff
@@ -229,6 +196,21 @@ def __call__(self, data):
         return data
 
 
+class PCACompute(object):
+    """
+    compute PCA of a point cloud and store the eigen values and the eigenvectors
```
nicolas-chaulet, line 201: Please add to the API doc and specify which attributes get added to the object `data`

src/core/data_transform/filters.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __call__(self, data):
+
+        res = True
+        for filter_fn in self.list_filter:
+            res = (res and filter_fn(data))
+        return res
+
+
+class PlanarityFilter(object):
```
nicolas-chaulet, line 20: Add to the doc, please also add the parameters of the `__init__` to the class docstring. You can take a look at the transforms in transforms.py as an example.

src/core/data_transform/filters.py
```diff
@@ -0,0 +1,44 @@
...
+    def __call__(self, data):
+        if(getattr(data, 'eigenvalues', None) is None):
+            data = PCACompute()(data)
+        planarity = compute_planarity(data.eigenvalues)
+        return planarity < self.thresh
+
+
+class RandomFilter(object):
+    """
+    Randomly select an elem or not (to have smaller dataset)
```
nicolas-chaulet, line 37: Same comment

test/test_msdatapair.py
```diff
@@ -0,0 +1,94 @@
+import unittest
+import torch
+import torch.testing as tt
+from torch_geometric.data import Data
+
+
+from src.datasets.registration.pair import Pair, MultiScalePair, PairBatch, PairMultiScaleBatch
+
+
+class TestMSPair(unittest.TestCase):
```
nicolas-chaulet, line 10: Nice!!

src/core/data_transform/filters.py
nicolas-chaulet: Also, could you please add a test for that filter?

src/core/data_transform/features.py
nicolas-chaulet: And please test! It is likely to be used and it would be annoying to have a bug in there...

src/core/data_transform/filters.py
humanpose1: For random Filter, I don't think it is necessary to test this class (too simple). However, I've tested FCompose (which combine different filters).

src/core/data_transform/features.py
humanpose1: I'm thinking it is more appropriate to add the PCACompute in torch-points-kernel because it is faster to do it in c++ (Here it's for one patch but if we want to do it for every patches...).
I still let PCACompute here, because I'm using it.

nicolas-chaulet: Yes, let's leave it here for now. We might add some c++ kernels to the repo itself as well, just for simplicity of development

src/core/data_transform/filters.py
nicolas-chaulet: Yes, thanks!

PR #217, opened 2020-04-08
Minkowski engine for registration

humanpose1:
I'm still testing it so There are still things to check. The network is slightly different from the original FCGF(I will so ). Can I have your feedback ?
I've skipped a test in test_models. I will try to deal with it later.


conf/data/registration/fragment3dmatch_sparse.yaml
```diff
@@ -0,0 +1,45 @@
...
+  dataroot: data
+  name: 3DMatch
+  is_patch: False
+  num_frame_per_fragment: 101
+  max_dist_overlap: 0.04
+  min_overlap_ratio: 0.2
+  tsdf_voxel_size: 0.006
+  limit_size: 850
+  depth_thresh: 4.5
+  first_subsampling: 0.02
```
nicolas-chaulet, line 13: You probably need only one of those two parameters

conf/data/registration/test3dmatch_sparse.yaml
```diff
@@ -0,0 +1,27 @@
+data:
+  task: registration
+  class: test3dmatch.Test3DMatchDataset
+  dataroot: data
+  name: 3DMatch
+  is_patch: False
+  radius_patch: 0.3
+  first_subsampling: 0.02
```
nicolas-chaulet, line 8: Same comment, you can keep only one

src/datasets/base_dataset.py
```diff
@@ -130,8 +129,7 @@ def set_filter(self, dataset_opt):
                     continue
                 setattr(self, new_name, filt)
 
-    @staticmethod
-    def _get_collate_function(conv_type, is_multiscale):
+    def _get_collate_function(self, conv_type, is_multiscale):
```
nicolas-chaulet, line 132: Why did you change that?

src/datasets/registration/base_siamese_dataset.py
```diff
@@ -14,9 +14,7 @@ def __init__(self, dataset_opt):
         """
         super().__init__(dataset_opt)
 
-
-    @staticmethod
-    def _get_collate_function(conv_type, is_multiscale):
+    def _get_collate_function(self, conv_type, is_multiscale):
```
nicolas-chaulet, line 17: Same as previous

test/test_models.py
```diff
@@ -114,6 +115,16 @@ def test_pointnet2ms(self):
         model.forward()
         model.backward()
 
+    @unittest.skip("The model is loading but Problem in the pairmockdataset...")
```
nicolas-chaulet, line 118: How come your model does not break the test `test_runall` above? It should break I think no?

src/models/base_architectures/unet.py
```diff
@@ -419,7 +419,7 @@ def _init_from_compact_format(self, opt, model_type, dataset, modules_lib):
             self.up_modules.append(up_module)
 
         self.metric_loss_module, self.miner_module = BaseModel.get_metric_loss_and_miner(
-            getattr(opt, "loss", None), getattr(opt, "miner", None)
+            getattr(opt, "metric_loss", None), getattr(opt, "miner", None)
```
nicolas-chaulet, line 422: Are you sure you want to change that? Have you made sure you've updated the rest of the configs?

src/models/registration/minkowski.py
```diff
@@ -0,0 +1,197 @@
...
+        self.loss_reg = self.metric_loss_module(self.output, self.labels, hard_pairs)
+        self.internal = self.get_internal_loss()
+        self.loss = self.loss_reg + self.internal
+
+    def forward(self):
+        self.output = self.apply_nn(self.input)
+        if self.labels is None:
+            return self.output
+        else:
+            self.apply_nn(self.input_target)
```
nicolas-chaulet, line 93: should be ` self.output_target = self.apply_nn(self.input_target)` I think

src/models/registration/minkowski.py
```diff
@@ -0,0 +1,197 @@
...
+            self.ind = data.pair_ind[rand_ind:, 0].to(device)
+            self.ind_target = data.pair_ind[rand_ind:, 1].to(device)
+            rang = torch.range(0, self.num_pos_pairs)
+            self.labels = torch.cat([rang, rang], 0).to(device)
+        else:
+            self.labels = None
+
+    def apply_nn(self, input):
+        output = self.model(self.input)
+        return ME.SparseTensor(
```
nicolas-chaulet, line 70: Do you need a sparse tensor here?

src/models/registration/minkowski.py
```diff
@@ -0,0 +1,197 @@
...
+            option.model_name, dataset.feature_dimension, option.out_channels, option.D
+        )
+        self.metric_loss_module, self.miner_module = BaseModel.get_metric_loss_and_miner(
+            getattr(option, "metric_loss", None), getattr(option, "miner", None)
+        )
+        self.num_pos_pairs = option.num_pos_pairs
+        self.loss_names = ["loss_reg", "loss", "internal"]
+
+        # Last Layer
+        last_mlp_opt = option.mlp_cls
```
nicolas-chaulet, line 27: You don't need all that here I believe, just remove it

src/models/registration/minkowski.py
```diff
@@ -0,0 +1,197 @@
...
+            return x
+
+    def compute_loss(self):
+        # miner
+
+        hard_pairs = None
+        if self.miner_module is not None:
+            hard_pairs = self.miner_module(self.output, self.labels)
+        # loss
+        self.loss_reg = self.metric_loss_module(self.output, self.labels, hard_pairs)
```
nicolas-chaulet, line 180: Why is this one named `loss_reg`? WHat does `reg` stand for?

src/modules/MinkowskiEngine/modules.py
```diff
@@ -376,3 +376,89 @@ class SEBottleneckIN(SEBottleneck):
...
+        kernel_size=3,
+        stride=1,
+        dilation=1,
+        has_bias=False,
+        activation=ME.MinkowskiReLU(inplace=True),
+        bn_momentum=0.01,
+        dimension=-1,
+        **kwargs
+    ):
+        """
```
nicolas-chaulet, line 438: Please put this doc_string (and others if there are) right after the class name, line 426. By doing so, an IDE gives you the docstring of the `__init__` method when you create a new instance of the class, quite handy.

src/datasets/base_dataset.py
humanpose1: I change this function because ii is apply in dataloaders[(Here)](https://github.com/nicolas-chaulet/deeppointcloud-benchmarks/blob/daed7c585c4acc45f38fee6a1ceb5ff5a6dd590e/src/datasets/base_dataset.py#L167) and the method didn't change it when I use it for a siamese_base_dataset class.

test/test_models.py
humanpose1: This is the only test that doesn't work. I got an error, with the pairmockdataset. But the model loads.

src/models/registration/minkowski.py
humanpose1: reg stands for registration... Maybe I should name it loss_desc. It is more appropriate.

src/models/registration/minkowski.py
humanpose1: Not explicitly.

src/models/registration/minkowski.py
humanpose1: Yes, (I didn't test this model that's why I haven't notice it !!). Thanks !

src/models/base_architectures/unet.py
humanpose1: Yes, I want to be specific with this loss which is associated with the miner. The loss and the miner are instantiated together for this reason. But we could imagine in the future adding different kind of losses for other things.

test/test_models.py
nicolas-chaulet: I am thinking that the run_all one passes because it actually does not run without a pair dataset. If that's the case then it would be good to adapt it so that all registration models do a forward and a backward (just like for segmentation).

src/models/registration/minkowski.py
nicolas-chaulet: Ok, makes sense. just keep it like that it's fine

humanpose1:
For now I'm still training Minkowski (it takes time...). There are still some problems I need to fix...
For now, I've tested Pointnet++ for the patch descriptors (Only [this paper](https://arxiv.org/pdf/1802.02669.pdf) mentionned Vanilla Pointnet for the patch registration). I have similar results(Which is not good...). That's why I prefer trying unet like architecture which is more promising.
The state of the art in patch is still [3DSmoothNet ](https://arxiv.org/pdf/1811.06879.pdf) but I haven' the intention to reproduce this method.

test/test_models.py
humanpose1: You're right. I'll do it !

src/datasets/base_dataset.py
```diff
@@ -171,7 +169,7 @@ def create_dataloaders(
         """
         conv_type = model.conv_type
         self._batch_size = batch_size
-        batch_collate_function = BaseDataset._get_collate_function(conv_type, precompute_multi_scale)
+        batch_collate_function = self._get_collate_function(conv_type, precompute_multi_scale)
```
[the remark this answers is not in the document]
humanpose1: I will show you an example:
```python
class A(object):

    @staticmethod
    def static(param1, param2):
        return param1*param2


    def other_function(self, param1, param2):
        print(A.static(param1, param2))


class B(A):
    @staticmethod
    def static(param1, param2):
        return param1**2 * param2**2


if __name__ == "__main__":

    a = A()

    a.other_function(3, 4) # 12

    b = B()
    b.other_function(3, 4) # 12
```

Here with non static method
```python
class A(object):

    def static(self, param1, param2):
        return param1*param2


    def other_function(self, param1, param2):
        print(self.static(param1, param2))


class B(A):
    def static(self, param1, param2):
        return param1**2 * param2**2


if __name__ == "__main__":

    a = A()

    a.other_function(3, 4) # 12

    b = B()
    b.other_function(3, 4) # 144
```
Ok I've found the solution.

```python
class A(object):

    @staticmethod
    def static(param1, param2):
        return param1*param2


    def other_function(self, param1, param2):
        print(self.__class__().static(param1, param2))


class B(A):
    @staticmethod
    def static(param1, param2):
        return param1**2 * param2**2


if __name__ == "__main__":

    a = A()

    a.other_function(3, 4) # 12

    b = B()
    b.other_function(3, 4) # 144
```

conf/models/registration/minkowski.yaml
```diff
@@ -0,0 +1,46 @@
+models:
+    # Minkowski Engine: https://github.com/StanfordVL/MinkowskiEngin
+    # STILL NOT FULLY WORKING !
```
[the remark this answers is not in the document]
humanpose1: Actually SimpleBlockUp and SimpleBlockDown are already residual layers (I should maybe change the name).

test/test_models.py
```diff
@@ -26,7 +29,7 @@ def load_model_config(task, model_type, model_name):
     models_conf = os.path.join(ROOT, "conf/models/{}/{}.yaml".format(task, model_type))
     config = OmegaConf.load(models_conf)
     config.update("model_name", model_name)
-    config.update("data.task", "segmentation")
+    config.update("data.task", task)
```
nicolas-chaulet, line 32: Awesome!!

test/test_models.py
```diff
@@ -55,13 +58,21 @@ def is_known_to_fail(model_name):
...
+                    return PairMockDataset(features, num_points=2048)
+                if conv_type.lower() == "sparse":
+                    tr = Compose([XYZFeature(True, True, True), ToSparseInput(0.01)])
+                    return PairMockDatasetGeometric(features, transform=tr, num_points=1024)
+                return PairMockDatasetGeometric(features)
+            else:
+                if conv_type.lower() == "dense":
+                    return MockDataset(features, num_points=2048)
+                if conv_type.lower() == "sparse":
+                    return MockDatasetGeometric(features, transform=ToSparseInput(0.01), num_points=1024)
```
nicolas-chaulet, line 74: For furture reference, ToSparseInput is actually a duplicate of the GridSampling with quantize=True. So maybe we could just use GridSampling accross the board.

torch_points3d/core/data_transform/feature_augment.py
```diff
@@ -133,3 +133,20 @@ def __call__(self, data):
 
     def __repr__(self):
         return "DropFeature: proba = {}, feature = {}".format(self._drop_proba, self._feature_name)
+
+
+class Jitter:
```
nicolas-chaulet, line 138: Please add to the docs :) and maybe rename to JitterFeature

torch_points3d/core/losses/metric_losses.py
```diff
@@ -0,0 +1,127 @@
...
+    if dist_type == "L2":
+        D2 = torch.sum((A.unsqueeze(1) - B.unsqueeze(0)).pow(2), 2)
+        return torch.sqrt(D2 + 1e-7)
+    elif dist_type == "SquareL2":
+        return torch.sum((A.unsqueeze(1) - B.unsqueeze(0)).pow(2), 2)
+    else:
+        raise NotImplementedError("Not implemented")
+
+
+class ContrastiveHardestNegativeLoss(nn.Module):
```
nicolas-chaulet, line 33: Could you add a reference here please?

torch_points3d/core/losses/metric_losses.py
```diff
@@ -0,0 +1,127 @@
...
+        return pos_loss.mean(), (neg_loss0.mean() + neg_loss1.mean()) / 2
+
+    def forward(self, F0, F1, matches, xyz0=None, xyz1=None):
+
+        pos_loss, neg_loss = self.contrastive_hardest_negative_loss(F0, F1, matches.detach().cpu())
+
+        return pos_loss + neg_loss
+
+
+class BatchHardContrastiveLoss(nn.Module):
```
nicolas-chaulet, line 97: Same a reference would be nice

torch_points3d/datasets/registration/test3dmatch.py
```diff
@@ -37,10 +37,15 @@ def __init__(self,
     def get(self, idx):
         data = torch.load(
             osp.join(self.path_table, 'fragment_{:06d}.pt'.format(idx)))
+        data.mask = torch.zeros(len(data.pos))
+        data.mask[data.keypoints] = 1
+        if(self.transform is not None):
```
nicolas-chaulet, line 42: This happens in the __get_item__ method already. Are you sure you need it there?

torch_points3d/models/registration/minkowski.py
```diff
@@ -0,0 +1,205 @@
...
+        self.loss_names = ["loss_reg", "loss"]
+        self.metric_loss_module, self.miner_module = BaseModel.get_metric_loss_and_miner(
+            getattr(option, "metric_loss", None), getattr(option, "miner", None)
+        )
+        # Last Layer
+
+        if option.mlp_cls is not None:
+            last_mlp_opt = option.mlp_cls
+            in_feat = last_mlp_opt.nn[0]
+            self.FC_layer = Sequential()
```
nicolas-chaulet, line 28: I added a new helper for simpler sequential models:
https://github.com/nicolas-chaulet/torch-points3d/blob/e32b2b2d1a1835f7e2ac9448d2d4e02ccfc22e48/torch_points3d/core/common_modules/base_modules.py#L156

scripts/test_registration_scripts/descriptor_matcher.py
```diff
@@ -63,6 +64,8 @@ def compute_dists(kp_source, kp_target, trans):
     """
     kp_target_t = kp_target.dot(trans[:3, :3].T) + trans[:3, 3]
     dist = np.linalg.norm(kp_source - kp_target_t, axis=1)
+    plt.hist(dist, bins=100)
```
tchaton, line 67: You might want to add plot_hist=False params here, so you don't plot the hist every time you call the function.

scripts/test_registration_scripts/descriptor_matcher.py
```diff
@@ -110,6 +118,7 @@ def pair_evaluation(path_descr_source, path_descr_target, gt_trans, list_tau, re
         name_source=n_s,
         name_target=n_t,
     )
+    print(n_s, n_t, frac_correct)
```
tchaton, line 121: please, use logger instead of print.

test/test_models.py
tchaton: Smart !

torch_points3d/core/data_transform/feature_augment.py
```diff
@@ -133,3 +133,20 @@ def __call__(self, data):
...
+    """
+    add a small gaussian noise to the feature
+    """
+    def __init__(self, mu=0, sigma=0.01):
+        self.mu = mu
+        self.sigma = sigma
+
+    def __call__(self, data):
+        if(random.random() < 0.95):
```
tchaton, line 147: 0.95 could be a p=0.95 argument

scripts/test_registration_scripts/descriptor_matcher.py
tchaton: Forget about this one

README.md
```diff
@@ -124,6 +124,11 @@ You are now in a position to install MinkowskiEngine with GPU support:
 poetry install -E MinkowskiEngine --no-root
 ```
 
+#### Pycuda
+```bash
+poetry add pycuda
```
nicolas-chaulet, line 129: Change for pip install pycuda. You don’t really want to add it to the poetry file

PR #218, opened 2020-04-09
S3disfullres + Precompute optimisation

nicolas-chaulet:
This is still work in progress but the idea is there.
TODO

- [x] test
- [x] run it on some pre trained models


eval.py
```diff
@@ -23,45 +23,79 @@
...
+    tracker_options={},
+):
     tracker.reset("val")
     loader = dataset.val_dataloader
-    with Ctq(loader) as tq_val_loader:
-        for data in tq_val_loader:
-            with torch.no_grad():
-                model.set_input(data, device)
-                model.forward()
+    for i in range(voting_runs):
```
tchaton, line 37: Awesome ! We have the voting runs in place :)

src/core/data_transform/transforms.py
```diff
@@ -61,6 +61,8 @@ class PointCloudFusion(object):
     """
 
     def _process(self, data_list):
+        if len(data_list) == 0:
+            return Data()
```
tchaton, line 65: Are you sure ? Won't it be better to raise an Exception there ?

src/core/data_transform/transforms.py
```diff
@@ -102,7 +104,7 @@ class GridSphereSampling(object):
     def __init__(self, radius, grid_size=None, delattr_kd_tree=True, center=True):
         self._radius = eval(radius) if isinstance(radius, str) else float(radius)
 
-        self._grid_sampling = GridSampling(size=grid_size if grid_size else self._radius)
+        self._grid_sampling = GridSampling(size=eval(grid_size) if grid_size else self._radius)
```
tchaton, line 107: Great catch !

src/core/data_transform/transforms.py
```diff
@@ -142,7 +144,7 @@ def _process(self, data):
                     item = item[ind]
                     if self._center and key == "pos":  # Center the sphere.
                         item -= t_center
-                    setattr(new_data, key, item)
+                setattr(new_data, key, item)
```
tchaton, line 147: Good catch there.

src/datasets/segmentation/s3dis.py
```diff
@@ -185,202 +186,6 @@ def get_tracker(model, dataset, wandb_log: bool, tensorboard_log: bool):
         return SegmentationTracker(dataset, wandb_log=wandb_log, use_tensorboard=tensorboard_log)
 
 
-################################### Used for s3dis radius sphere ###################################
-
-
-class S3DISOriginal(InMemoryDataset):
```
tchaton, line 191: Ca fait mal au coeur ca ahaha

src/datasets/segmentation/s3dis.py
```diff
@@ -500,6 +301,10 @@ def process(self):
                 if os.path.isdir(osp.join(self.raw_dir, f, room_name))
             ]
 
+            train_files = train_files[1:10]
```
tchaton, line 304: You might want to remove [1:10]

src/metrics/s3dis_tracker.py
```diff
@@ -0,0 +1,107 @@
...
+        # Train mode or low res, nothing special to do
+        if self._stage == "train" or not full_res:
+            return
+
+        # Test mode, compute votes in order to get full res predictions
+        if self._test_area is None:
+            self._test_area = self._dataset.test_data.clone()
+            if self._test_area.y is None:
+                raise ValueError("It seems that the test area data does not have labels (attribute y).")
+            self._test_area.has_prediction = torch.zeros(self._test_area.y.shape[0], dtype=torch.bool)
```
tchaton, line 39: Brilliant !

src/metrics/s3dis_tracker.py
```diff
@@ -0,0 +1,107 @@
...
+        t = time.time()
+        full_pred = knn_interpolate(
+            self._test_area.votes[self._test_area.has_prediction],
+            self._test_area.pos[self._test_area.has_prediction],
+            self._test_area.pos,
+            k=1,
+        )
+        print("knn timing: %.2f" % (time.time() - t))
+
+        # Complete for points that have a prediction
```
tchaton, line 82: Why computing low res in compute_full_miou ? Please, do two functions which can be both activated if needed.

src/metrics/segmentation_tracker.py
```diff
@@ -5,10 +5,13 @@
 from src.models.base_model import BaseModel
 from src.metrics.confusion_matrix import ConfusionMatrix
 from src.metrics.base_tracker import BaseTracker, meter_value
+from src.datasets.segmentation import IGNORE_LABEL
```
tchaton, line 8: Great to have it in segmentation :)

src/core/data_transform/transforms.py
nicolas-chaulet: I think it's ok, it is happening on on degenerated and if you try to do something useful with what it returns you'll quickly see that something is wrong (data is empty)

src/datasets/segmentation/s3dis.py
nicolas-chaulet: Yes, we were not using it... And it was actually broken in a few places. So thought I would just remove it

src/datasets/segmentation/s3dis.py
nicolas-chaulet: Yes done!

src/metrics/s3dis_tracker.py
nicolas-chaulet: Good point, it is for debugging for now. I will revisit once all works

eval.py
nicolas-chaulet: Yes! The tracker needs to be able to do something useful with it (like s3dis) but in principle it's there :)

src/metrics/s3dis_tracker.py
```diff
@@ -0,0 +1,107 @@
...
+        if inputs[SaveOriginalPosId.KEY] is None or inputs[SaveOriginalPosId.KEY].max() >= self._test_area.pos.shape[0]:
+            raise ValueError(
+                "The inputs given to the model do not have a %s attribute or this attribute does\
+                     not correspond to the number of points in the test area point cloud."
+                % SaveOriginalPosId.KEY
+            )
+
+        # Set predictions
+        outputs = model.get_output()
+        self._test_area.votes[inputs[SaveOriginalPosId.KEY]] += outputs
```
tchaton, line 54: You should normalize the probs with a counter if knn interpolate uses k > 1. (You don't have probs anymore)

nicolas-chaulet: Will do

torch_points3d/core/data_transform/transforms.py
```diff
@@ -424,14 +424,15 @@ def __call__(self, data: Data) -> MultiScaleData:
         # Compute sequentially multi_scale indexes on cpu
         data.contiguous()
         ms_data = MultiScaleData.from_data(data)
-        precomputed = [data]
+        precomputed = [Data(pos=data.pos)]
```
tchaton, line 427: Why is it needed here ?

torch_points3d/datasets/multiscale_data.py
```diff
@@ -124,7 +125,7 @@ def from_data_list_token(data_list, follow_batch=[]):
     for i, data in enumerate(data_list):
         for key in data.keys:
             item = data[key]
-            if torch.is_tensor(item) and item.dtype != torch.bool:
+            if torch.is_tensor(item) and item.dtype != torch.bool and cumsum[key] > 0:
```
tchaton, line 128: Why cumsum[key] > 0 ?

torch_points3d/metrics/s3dis_tracker.py
```diff
@@ -0,0 +1,113 @@
...
+            raise ValueError(
+                "The inputs given to the model do not have a %s attribute or this attribute does\
+                     not correspond to the number of points in the test area point cloud."
+                % SaveOriginalPosId.KEY
+            )
+
+        # Set predictions
+        outputs = model.get_output()
+        self._test_area.votes[inputs[SaveOriginalPosId.KEY]] += outputs
+        self._test_area.has_prediction[inputs[SaveOriginalPosId.KEY]] = True
```
tchaton, line 55: Might be better to track the count and then use it to get the mask as > 0

torch_points3d/core/data_transform/transforms.py
nicolas-chaulet: This makes sure that only pos go though the multiscale, we don't need the other fields in data

torch_points3d/metrics/s3dis_tracker.py
nicolas-chaulet: Yes

torch_points3d/datasets/multiscale_data.py
nicolas-chaulet: Makes sure we don't do the mask + select if we don't need to. It is quite slow.

PR #219, opened 2020-04-10
'add scannet from labels directly'

tchaton:
ElasticDistorsion + RemoveDuplicates is pretty slow. Need to investigate that.


src/core/data_transform/grid_transform.py
```diff
@@ -46,35 +90,27 @@ class GridSampling:
...
-
-        return group_data(data, cluster, perm, mode="mean")
+        coords = ((data.pos) / self._grid_size).int()
+        batch = data.batch if hasattr(data, "batch") else None
+        cluster, unique_pos_indices = sparse_coords_to_clusters(coords, batch)
+        if self._to_sparse_coords:
+            delattr(data, "pos")
+        data = group_data(data, cluster, unique_pos_indices, mode="mean")
+        if self._to_sparse_coords:
+            data.pos = coords[unique_pos_indices]
```
humanpose1, line 85: It could also be great to store the unique_pos_indices.

src/metrics/meters.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+
+    def value(self):
+        """Get the value of the meter in the current state."""
+
+
+# This code has been taken from Pytorch Torchnet has it contains a bug with an assert
+# https://github.com/pytorch/tnt/issues/131
+class APMeter(Meter):
+    """
```
nicolas-chaulet, line 26: Nice addition!

src/models/base_model.py
```diff
@@ -153,11 +154,13 @@ def optimize_parameters(self, epoch, batch_size):
...
 
+        scheduler_step = epoch if self._update_scheduler_on_epoch else self._iterations
+
         if self._lr_scheduler:
-            self._lr_scheduler.step(epoch)
+            self._lr_scheduler.step(scheduler_step)
 
         if self._bn_scheduler:
-            self._bn_scheduler.step(epoch)
+            self._bn_scheduler.step(scheduler_step)
```
nicolas-chaulet, line 163: Could we test that? I think there is already a test fro the schedulers, would be nice to test both modes

src/core/data_transform/features.py
```diff
@@ -20,6 +21,59 @@
 from src.utils.config import is_list
 from src.utils import is_iterable
 
+class RandomRotation(object):
```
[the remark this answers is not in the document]
tchaton: I find only Random Rotate, but there were no transform doing a 3 random transform at once.

src/core/data_transform/sparse_transforms.py
```diff
@@ -200,4 +199,65 @@ def __call__(self, data):
...
         return "{}(grid_size={}, save_delta={}, save_delta_norm={}, mode={})"\
-            .format(self.__class__.__name__, self._grid_size, self._save_delta, self._save_delta_norm, self._mode)
\ No newline at end of file
+            .format(self.__class__.__name__, self._grid_size, self._save_delta, self._save_delta_norm, self._mode)
+
+
+
+class ElasticDistortion:
+
+    """Apply elastic distortion on sparse coordinate space.
```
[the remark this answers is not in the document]
tchaton: Not sure what you mean ?

src/datasets/segmentation/scannet.py
```diff
@@ -672,3 +679,266 @@ def get_tracker(model, dataset, wandb_log: bool, tensorboard_log: bool):
...
+                log.info("SAVING TO {}".format(self.processed_paths[i]))
+                torch.save(self.collate(datas), self.processed_paths[i])
+        log.info("SAVING TO {}".format(self.processed_paths[-1]))
+        torch.save(self.collate(all_datas), self.processed_paths[-1])        
+
+    def __repr__(self):
+        return "{}({})".format(self.__class__.__name__, len(self))
+
+
+class ScannetDatasetPreComputed(BaseDataset):
```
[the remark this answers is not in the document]
tchaton: The one above is doing boxes regression, instance segmentation, segmentation.
The one used for Minkowski Engine is the second one, which should be exactly the same as the above segmentation once.

However, I preferred to follow what they did to minimise chance of bugs.

src/models/base_model.py
tchaton: Yes, definitely.

src/core/data_transform/features.py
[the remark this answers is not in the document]
nicolas-chaulet: Indeed it does only support one axis. I know we have this pattern in other places, are you sure you want to have a parameter that disables the transform? Removing it from the config file seems cleaner no? I would never imagine that someone has put a transform in a config file but this transform is inactive. It is the same with add features by key, if you don't pay attention you think all features listed are appended but it could be that one is actually not added which requires extra attention when reading the config. Is the use case so that you can control that through command line?

src/core/data_transform/sparse_transforms.py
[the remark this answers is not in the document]
nicolas-chaulet: The docstring should be straight after the class name, you have a line jump there that should not be there. It's weird that black did not pick that up though

src/datasets/segmentation/scannet.py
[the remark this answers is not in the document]
nicolas-chaulet: Makes sense. Would it be possible to consolidate both before we merge the PR?

src/core/data_transform/grid_transform.py
```diff
@@ -48,33 +85,47 @@ class GridSampling:
...
+        
+        coords = ((data.pos) / self._grid_size).int()
+        if self._elastic_distorsion:
+            coords = self._distorsion(Data(pos=coords)).pos
+        batch = data.batch if hasattr(data, "batch") else None
+        cluster, unique_pos_indices = sparse_coords_to_clusters(coords, batch)
+        
+        # Delete pos for small speed up
+        if self._quantize_coords:
+            delattr(data, "pos")
```
humanpose1, line 123: Actually, it's a bad idea. especially if x is still empty. You should remove this line because num_nodes will be None if only pos is in data.

src/core/data_transform/features.py
```diff
@@ -14,13 +14,47 @@
...
+
+    def __init__(self, apply_rotation:bool = True, rot_x: float = None, rot_y: float = None, rot_z: float = None):
+        self._apply_rotation = apply_rotation
+        self._rot_x = rot_x
+        self._rot_y = rot_y
+        self._rot_z = rot_z
+
+        from torch_geometric.transforms import Compose, RandomRotate
+
+        self.rotations = Compose([RandomRotate(rot_x, axis=0), RandomRotate(rot_y, axis=1), RandomRotate(rot_z, axis=2)])
```
nicolas-chaulet, line 47: Nice

src/core/data_transform/grid_transform.py
nicolas-chaulet: @tchaton he has a good point could you please correct?

src/core/data_transform/grid_transform.py
```diff
@@ -97,3 +149,61 @@ class SaveOriginalPosId:
     def __call__(self, data):
         setattr(data, self.KEY, torch.arange(0, data.pos.shape[0]))
         return data
+
+class ElasticDistortion:
+    """Apply elastic distortion on sparse coordinate space.
```
nicolas-chaulet, line 154: Could you please point to something that explains what this does? At the moment the doc does not help much.

src/core/data_transform/grid_transform.py
```diff
@@ -97,3 +149,61 @@ class SaveOriginalPosId:
...
+
+    Parameters
+    ----------
+    granularity: float
```
nicolas-chaulet, line 158: Parameters don't match with the actual parameters in init

src/core/data_transform/features.py
```diff
@@ -25,39 +25,65 @@ class Random3AxisRotation(object):
...
+    @staticmethod
+    def rotation_matrix(axis, theta):
+        return expm(np.cross(np.eye(3), axis / norm(axis) * theta))
 
     def __call__(self, data):
         if self._apply_rotation:
-            data = self.rotations(data)
+            pos = data.pos
+            M = self.generate_random_rotation_matrix()
+            data.pos = M @ pos
```
humanpose1, line 82: It's weird. Normally pos is size N x 3 and M should be of size 3 x 3. Isn't there a problem of broadcast ?

test/test_config/lr_scheduler.yaml
```diff
@@ -0,0 +1,44 @@
...
+        # accumulated_gradient: -1 # Accumulate gradient accumulated_gradient * batch_size
+        grad_clip: -1
+        optimizer:
+            class: SGD
+            params:
+                lr: ${training.optim.base_lr} # The path is cut from training
+                momentum: 0.9
+                dampening: 0.1
+                weight_decay: 1
+        lr_scheduler:
```
humanpose1, line 19: I like the idea !!!

src/core/data_transform/features.py
```diff
@@ -25,39 +25,65 @@ class Random3AxisRotation(object):
...
-        self._rot_z = rot_z
-
-        from torch_geometric.transforms import Compose, RandomRotate
+        self._rot_x = np.abs(rot_x) if rot_x else rot_x
+        self._rot_y = np.abs(rot_y) if rot_y else rot_y
+        self._rot_z = np.abs(rot_z) if rot_z else rot_z
+
+        self._degree_angles = [self._rot_x, self._rot_y, self._rot_z]
+
+    def generate_random_rotation_matrix(self):
```
humanpose1, line 50: Why you don't use the function euler_angle_to_rotation_matrix ?

tchaton: Where is this function ?

humanpose1:
```python
def euler_angles_to_rotation_matrix(theta):
    R_x = torch.tensor(
        [[1, 0, 0], [0, torch.cos(theta[0]), -torch.sin(theta[0])], [0, torch.sin(theta[0]), torch.cos(theta[0])]]
    )

    R_y = torch.tensor(
        [[torch.cos(theta[1]), 0, torch.sin(theta[1])], [0, 1, 0], [-torch.sin(theta[1]), 0, torch.cos(theta[1])]]
    )

    R_z = torch.tensor(
        [[torch.cos(theta[2]), -torch.sin(theta[2]), 0], [torch.sin(theta[2]), torch.cos(theta[2]), 0], [0, 0, 1]]
    )

    R = torch.mm(R_z, torch.mm(R_y, R_x))
    return R
```

src/core/data_transform/feature_augment.py
```diff
@@ -3,6 +3,25 @@
 
 # Those Transformation are adapted from https://github.com/chrischoy/SpatioTemporalSegmentation/blob/master/lib/transforms.py
 
+class NormalizeRGB(object):
```
nicolas-chaulet, line 6: It's already there

src/core/data_transform/features.py
```diff
@@ -25,39 +25,69 @@ class Random3AxisRotation(object):
...
     Rotate pointcloud with random angles along x, y, z axis
 
+    The angles should be given into degree.
+
     Parameters
     -----------
     apply_rotation: bool:
         Whether to apply the rotation
     rot_x: float
-        Rotation angle on x axis in degrees
```
nicolas-chaulet, line 33: Add the in degrees mention back. It is not very standard

src/core/data_transform/features.py
```diff
@@ -25,39 +25,69 @@ class Random3AxisRotation(object):
...
         )
 
+        R = torch.mm(R_z, torch.mm(R_y, R_x))
+        return R
+
+    def generate_random_rotation_matrix(self):
+        thetas = []
+        for axis_ind, deg_angle in enumerate(self._degree_angles):
+            if deg_angle > 0:
+                rand_deg_angle = np.random.uniform(0, deg_angle)
```
nicolas-chaulet, line 75: Use random.random instead of numpy random. Numpy random is initialised with the same see at each epoch for so strange reason. So you end up with the same samples all the time.

conf/data/segmentation/scannet.yaml
```diff
@@ -26,32 +29,41 @@ data:
...
+          apply_rotation: ${data.apply_rotation}
+          rot_x: 5
+          rot_y: 5
+          rot_z: 180
+      - transform: GridSampling
+        params:
+          size: ${data.grid_size}
+          quantize_coords: True
+          mode: "last"
+          elastic_distorsion: False
```
nicolas-chaulet, line 43: Those are not valid arguments anymore

src/core/schedulers/bn_schedulers.py
```diff
@@ -1,19 +1,22 @@
...
 def set_bn_momentum_default(bn_momentum):
     def fn(m):
-        if isinstance(m, (nn.BatchNorm1d, nn.BatchNorm2d, nn.BatchNorm3d)):
+        if isinstance(m, BATCH_NORM_MODULES):
             m.momentum = bn_momentum
 
     return fn
 
 
 class BNMomentumScheduler(object):
```
nicolas-chaulet, line 18: A docstring would be nice to explicit each parameter

src/datasets/segmentation/scannet.py
```diff
@@ -591,6 +603,16 @@ def process_func(
 
     def process(self):
         self.read_from_metadata()
+
+        mapping_dict = {indice:idx for idx, indice in enumerate(self.valid_class_idx)}
```
nicolas-chaulet, line 607: What is the benefit of this `mapping_func`? It is a bit harder to understand the code so would be good to have benefits from the change

src/metrics/model_checkpoint.py
```diff
@@ -44,10 +45,9 @@ def save_objects(self, models_to_save, stage, current_stat, optimizer, scheduler
         self.models = models_to_save
         self.optimizer = [optimizer.__class__.__name__, optimizer.state_dict()]
         self.schedulers = {
-            scheduler_name: [scheduler.scheduler_opt, scheduler.state_dict()]
+            scheduler_name: [scheduler.scheduler_opt, scheduler.state_dict(), scheduler.update_scheduler_on]
```
nicolas-chaulet, line 48: Could we add `scheduler.update_scheduler_on` to `scheduler.scheduler_opt` if it's not already there? It is a scheduler option at the end of the day no? This will also allow you to make the code retrocompatible

src/metrics/segmentation_tracker.py
```diff
@@ -1,10 +1,13 @@
 from typing import Dict
 import torchnet as tnt
+import torch
 import numpy as np
-
+from torch.nn import functional as F
 from src.models.base_model import BaseModel
 from src.metrics.confusion_matrix import ConfusionMatrix
 from src.metrics.base_tracker import BaseTracker, meter_value
+from src.utils.np_utils import to_categorical
```
nicolas-chaulet, line 9: Is this sued somewhere? I can't see it in this file. If not then please remove the whole file please

test/test_transform.py
```diff
@@ -166,13 +167,6 @@ def test_RemoveAttributes(self):
         for key in keys[mask]:
             self.assertNotIn(key, list(data_out.keys))
 
-    def test_RemoveDuplicateCoords(self):
```
nicolas-chaulet, line 169: Please add this test back

PR #225, opened 2020-04-18
Profiling support + cosmetic in transforms

nicolas-chaulet:
This is part of @tchaton PR, porting a few nice things without bringing the experimental stuff


src/core/data_transform/grid_transform.py
```diff
@@ -48,33 +85,43 @@ class GridSampling:
...
-    def __init__(self, size, start=None, end=None, num_classes=-1):
-        self.size = size
-        self.start = start
-        self.end = end
-        self.num_classes = num_classes
+    def __init__(self, size, quantize_coords=False, mode="mean"):
+        self._grid_size = size
+        self._quantize_coords = quantize_coords
+        self._mode = mode
+        log.warning(
```
humanpose1, line 100: Actually that's great to have this warning. But I think we should add a param warn  which has a default value. It will print this message only if warn is True, so the user can choose whether we print the msg or not.

src/core/data_transform/grid_transform.py
```diff
@@ -48,33 +85,43 @@ class GridSampling:
...
+            "If you need to keep track of the position of your points, use SaveOriginalPosId transform before using GridSampling"
+        )
+
+        if self._mode == "last":
+            log.warning(
```
humanpose1, line 105: same as above

src/core/data_transform/grid_transform.py
```diff
@@ -84,7 +131,9 @@ def __call__(self, data):
         return data
 
     def __repr__(self):
-        return "{}(size={})".format(self.__class__.__name__, self.size)
+        return "{}(grid_size={}, quantize_coords={}, mode={})".format(
+            self.__class__.__name__, self._grid_size, self._quantize_coords, self._mode
+        )
 
 
 class SaveOriginalPosId:
```
humanpose1, line 139: Maybe this class needs also a __repr__.

src/core/data_transform/grid_transform.py
tchaton: Yes. I am getting tired of seeing this while running unitest

src/core/data_transform/grid_transform.py
```diff
@@ -1,12 +1,42 @@
...
+    num_points = data.pos.shape[0]
+    shuffle_idx = torch.randperm(num_points)
+    for key in set(data.keys):
+        item = data[key]
+        if num_points == item.shape[0]:
+            data[key] = item[shuffle_idx]
+    return data
+
+
+def sparse_coords_to_clusters(pos, batch):
```
tchaton, line 22: Great PR !

PR #226, opened 2020-04-18
Faster grid

nicolas-chaulet:
Previous merge was really slow, this fixes it and gets us back to where we were


src/core/data_transform/grid_transform.py
```diff
@@ -14,28 +15,11 @@ def shuffle_data(data):
     shuffle_idx = torch.randperm(num_points)
     for key in set(data.keys):
         item = data[key]
-        if num_points == item.shape[0]:
+        if torch.is_tensor(item) and num_points == item.shape[0]:
```
tchaton, line 18: Great catch !

PR #227, opened 2020-04-19
Conversion into package 

tchaton:
python setup.py install

import torch_points3d

debug: I had to copy mockdatasets within utils. I didn't find a way to make it work without for now.


docs/index.rst
```diff
@@ -90,15 +90,15 @@ Supported tasks
    :caption: Contents
    :hidden:
 
-   src/gettingstarted
-   src/tutorials
-   src/advanced
+   torch_points3d/gettingstarted
```
nicolas-chaulet, line 93: Wrong path and the two after as well

docs/index.rst
```diff
@@ -90,15 +90,15 @@ Supported tasks
...
 
 .. toctree::
    :glob:
    :maxdepth: 1
    :caption: API
    :hidden:
 
-   src/api/transforms
-   src/api/filters
+   torch_points3d/api/transforms
```
nicolas-chaulet, line 103: Wrong path

torch_points3d/metrics/registration_accuracy.py
```diff
@@ -0,0 +1,23 @@
+import numpy as np
```
nicolas-chaulet, line 1: This file has been deleted in the previous merge

PR #235, opened 2020-04-21
repare XYZFeatures

tchaton:
This new XYZFeature add to an x, y, z attribute within data, not to x directly anymore
@humanpose1 update your code accordingly :)


nicolas-chaulet:
Nice!

PR #242, opened 2020-04-26
Votenet integration

tchaton:
This PR add VoteNet to the repo. 
It is training, but the metric aren't all implemented.


torch_points3d/modules/VoteNet/loss_helper.py
```diff
@@ -0,0 +1,325 @@
...
+import torch.nn as nn
+import numpy as np
+
+FAR_THRESHOLD = 0.6
+NEAR_THRESHOLD = 0.3
+GT_VOTE_FACTOR = 3  # number of GT votes per point
+OBJECTNESS_CLS_WEIGHTS = [0.2, 0.8]  # put larger weights on positive objectness
+
+
+def huber_loss(error, delta=1.0):
```
humanpose1, line 16: Can you put the loss as a torch module ?

torch_points3d/core/base_conv/dense.py
```diff
@@ -64,7 +65,10 @@ def forward(self, data, **kwargs):
             pos -- Previous positions [B, N, 3]
         """
         x, pos = data.x, data.pos
-        idx = self.sampler(pos)
+        if "idx" in data:
```
nicolas-chaulet, line 68: Sure about that?

torch_points3d/core/base_conv/dense.py
```diff
@@ -73,7 +77,11 @@ def forward(self, data, **kwargs):
             radius_idx = self.neighbour_finder(pos, new_pos, scale_idx=scale_idx)
             ms_x.append(self.conv(x, pos, new_pos, radius_idx, scale_idx))
         new_x = torch.cat(ms_x, 1)
-        return Data(pos=new_pos, x=new_x)
+
+        new_data = Data(pos=new_pos, x=new_x)
```
nicolas-chaulet, line 81: Probably not good to delete all the other attributes. Don't you want to keep sample ids?

torch_points3d/models/object_detection/votenet.py
```diff
@@ -0,0 +1,89 @@
...
+        - (required) call the initialization function of BaseModel
+        - define loss function, visualization images, model names, and optimizers
+        """
+        super(VoteNetModel, self).__init__(option)
+
+        # 1 - CREATE BACKBONE MODEL
+        input_nc = dataset.feature_dimension
+
+        backbone_option = option.backbone
+        self.num_point = backbone_option.down_conv.npoint[1]
```
nicolas-chaulet, line 28: I don't think you need it here. You only need it after the forward of the backbone and at that point the number of point is easily obtained by looking at the shape of the output of the forward. By using that you keep the possibility of having different architectures (larger or smaller encoders). This will also be important if we want to use grid sampling as our sampling strategy for example for KPConv backbone.

torch_points3d/models/object_detection/votenet.py
```diff
@@ -0,0 +1,89 @@
...
+        voting_option = option.voting
+        voting_cls = getattr(votenet_module, voting_option.module_name)
+        self.voting_module = voting_cls(vote_factor=voting_option.vote_factor, seed_feature_dim=voting_option.feat_dim)
+
+        # 3 - CREATE PROPOSAL MODULE
+        proposal_option = option.proposal
+        proposal_cls = getattr(votenet_module, proposal_option.module_name)
+        self.proposal_cls_module = proposal_cls(
+            num_class=proposal_option.num_class,
+            vote_aggregation_config=proposal_option.vote_aggregation,
```
nicolas-chaulet, line 42: When possible keep the naming in the config and in the functions the same (vote_aggregation or vote_aggregation_config, you can choose either one of them)

torch_points3d/models/object_detection/votenet.py
```diff
@@ -0,0 +1,89 @@
...
+        self.losses_has_been_added = False
+        self.loss_names = []
+
+    def set_input(self, data, device):
+        """Unpack input data from the dataloader and perform necessary pre-processing steps.
+        Parameters:
+            input: a dictionary that contains the data itself and its metadata information.
+        """
+        # Forward through backbone model
+        self.input = data
```
nicolas-chaulet, line 64: data.to(device)?

torch_points3d/models/object_detection/votenet.py
```diff
@@ -0,0 +1,89 @@
...
+        outputs = self.proposal_cls_module(data_votes)
+
+        sampling_id_key = "sampling_id_0"
+        setattr(outputs, "seed_inds", getattr(data_features, sampling_id_key, None)[:, : self.num_point,])
+
+        self.output = outputs
+
+    def backward(self):
+        """Calculate losses, gradients, and update network weights; called in every training iteration"""
+        losses, metrics, labels = votenet_module.get_loss(self.input, self.output, self.loss_params)
```
nicolas-chaulet, line 80: We usually compute the losses in the forward. Any reason why you have switched to computing them in the backward? This will break the eval loops for example

torch_points3d/models/object_detection/votenet.py
```diff
@@ -80,7 +82,7 @@ def _compute_losses(self):
             if torch.is_tensor(loss):
                 if not self.losses_has_been_added:
                     self.loss_names += [loss_name]
-                setattr(self, loss_name, loss.item())
+                setattr(self, loss_name, loss)
```
tchaton, line 85: Sure ? It is a tensor

nicolas-chaulet: Yes!

conf/data/object_detection/scannet.yaml
```diff
@@ -1,6 +1,6 @@
 data:
   class: scannet.ScannetDataset
-  dataset_name: "scannet"
+  dataset_name: "scannet-fixed"
```
tchaton, line 3: scannet-fixed should match file name.

PR #244, opened 2020-04-27
Removing Erik code

nicolas-chaulet:
Some more cleanup

- Remove Erik's code for convs and sequential models
- Simplify API by removing the call to the set_input method

### DIFF
--- a/torch_points3d/core/base_conv/dense.py
+++ b/torch_points3d/core/base_conv/dense.py
@@ -18,11 +18,10 @@ from torch_geometric.nn import (
 )
 from torch_geometric.data import Data
 import torch_points_kernels as tp
-import etw_pytorch_utils as pt_utils
 
 from torch_points3d.core.spatial_ops import BaseMSNeighbourFinder
 from torch_points3d.core.base_conv import BaseConvolution
-from torch_points3d.core.common_modules import MLP
+from torch_points3d.core.common_modules.dense_modules import MLP2D
 
 from torch_points3d.utils.enums import ConvolutionFormat
 from torch_points3d.utils.model_building_utils.activation_resolver import get_activation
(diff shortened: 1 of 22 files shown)


torch_points3d/core/base_conv/dense.py
```diff
@@ -138,9 +137,9 @@ def __repr__(self):
 
 
 class GlobalDenseBaseModule(torch.nn.Module):
-    def __init__(self, nn, aggr="max", bn=True, activation="LeakyReLU", **kwargs):
+    def __init__(self, nn, aggr="max", bn=True, activation=torch.nn.LeakyReLU(negative_slope=0.1), **kwargs):
         super(GlobalDenseBaseModule, self).__init__()
-        self.nn = pt_utils.SharedMLP(nn, bn=bn, activation=get_activation(activation))
+        self.nn = MLP2D(nn, bn=bn, activation=activation, bias=False)
```
tchaton, line 142: Why removing get_activation ? It was doing automatic resolution.

torch_points3d/core/common_modules/base_modules.py
```diff
@@ -132,11 +128,32 @@ def forward(self, features, category_labels, **kwargs):
...
     def __init__(self, num_features, momentum=0.1):
         super().__init__()
-        self.batch_norm = BN(num_features, momentum=momentum)
+        self.batch_norm = nn.BatchNorm1d(num_features, momentum=momentum)
 
-    def forward(self, x):
+    def _forward_dense(self, x):
+        return self.batch_norm(x)
+
+    def _forward_sparse(self, x):
```
tchaton, line 136: Small explanation here will be great for other people.

nicolas-chaulet: Done

torch_points3d/core/base_conv/dense.py
nicolas-chaulet: We were not using it. You can still change the activation by giving the full torch module

PR #252, opened 2020-05-02
update for pypi

tchaton:


pyproject.toml
```diff
@@ -22,6 +24,8 @@ torch-cluster = "==1.4.5"
 torch-geometric = "==1.3.2"
 pytorch_metric_learning = "^0.9.75"
 sphinx_rtd_theme = "^0.4.3"
+sphinx = "^2.4.4"
```
nicolas-chaulet, line 27: All Sphinx dependencies should go in dev. You need to poetry add and poetry remove them so that it works

.github/workflows/pythonapp.yml
```diff
@@ -32,6 +32,11 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: $HOME/.poetry/bin/poetry install
 
+      - name: Linting
+        run: |
```
nicolas-chaulet, line 36: Why did you add this? This will break if someone does not have pre commit installed. I’m ok with it, just want to make sure we’re aligned on what it does

tchaton: I was thinking so poetry build will always receive styled code. So if someone install it and look at the source, it is always formatted.
However, I am having some trouble, and I removed it.

PR #258, opened 2020-05-06
Clear intergface for models

nicolas-chaulet:
As part of the effort towards a release as a standalone package, I wanted to make the code more modular. For example it is a bot tricky today to use the trackers since there not documentation of what the interface of the model should be. This PR tries to bring a bit more clarity into that.


torch_points3d/models/model_interface.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class TrackerInterface(ABC):
+    @abstractmethod
+    def get_labels(self):
+        """ returns a trensor of size ``[N_points]`` where each value is the label of a point
+        """
+
+    @abstractmethod
+    def get_batch_idx(self):
```
humanpose1, line 43: I use the the name get_batch

torch_points3d/models/model_interface.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+
+    @abstractmethod
+    def get_output(self):
+        """ returns a trensor of size ``[N_points,...]`` where each value is the output
+        of the network for a point (output of the last layer in general)
+        """
+
+    @abstractmethod
+    def get_input(self):
```
humanpose1, line 54: I use the name get_xyz

torch_points3d/datasets/base_dataset.py
```diff
@@ -7,6 +7,7 @@
 import torch_geometric
 from torch_geometric.transforms import Compose, FixedPoints
 
+from torch_points3d.models.model_interface import DatasetInterface
```
tchaton, line 10: why modelInterface contains DatasetInterface ?

tchaton: Ok. model_interface name seems a little ambiguos. Maybe we could find a better way to name it, as external_interfaces_to_model, or something.
or regrouping all of them within an torch_points3d.core.interfaces

torch_points3d/models/model_interface.py
nicolas-chaulet: Get input could potentially contain the whole input data structure, it’s up to you

torch_points3d/models/model_interface.py
nicolas-chaulet: Get batch is better! I’ll change that

torch_points3d/datasets/base_dataset.py
nicolas-chaulet: Yes, not sure what’s the best way to name those things.

nicolas-chaulet: Another way to deal with it is to import model_interface and then use model_interface.DatasetInterface and so on. What do you think? In at case you need to reference model and dataset

nicolas-chaulet: If they are all in the same file it will be even more confusing I think

tchaton: Ok, let's keep it that way for now.

PR #259, opened 2020-05-07
Cleancode

nicolas-chaulet:
I have added mypy for type checking. It is useful to force us to stick to interfaces. We don't need to annotate everything but it is particularly important for "public" parts of our code base. To run locally simpy run `mypy torch_points3d`. @humanpose1 I have removed checks on the bits in the registration tracker. If/when you fix those please remove the `# type: ignore` comment that are in the tracker


test/test_basemodel.py
```diff
@@ -1,7 +1,7 @@
 import unittest
 from omegaconf import DictConfig
 from torch.nn import (
-    Sequential as Seq,
+    Sequential as Sequential,
```
tchaton, line 4: Don't need Sequential as Sequential, just Sequential

torch_points3d/metrics/registration_tracker.py
```diff
@@ -84,9 +84,9 @@ def track(self, model: model_interface.TrackerInterface, **kwargs):
         super().track(model)
         if self._stage != "train":
             batch_idx, batch_idx_target = model.get_batch()
-            batch_xyz, batch_xyz_target = model.get_xyz()
-            batch_ind, batch_ind_target, batch_size_ind = model.get_ind()
-            batch_feat, batch_feat_target = model.get_outputs()
+            batch_xyz, batch_xyz_target = model.get_xyz()  # type: ignore
+            batch_ind, batch_ind_target, batch_size_ind = model.get_ind()  # type: ignore
+            batch_feat, batch_feat_target = model.get_output()
```
humanpose1, line 89: If I understand well I should add static Typing to these functions right ?

nicolas-chaulet: Not necessarily, the problem is that they are not part of the main TrackerInterface. You have two (three) options:

- use the get_output and get_input methods to access what you need
- subclass the interface to add those methods
 - leave as is

I would prefer option 1 if that's doable

humanpose1: the problem is get_xyz depends on the networks. If it's kpconv or minkowski it will not be the same function.  the option 2 seems more appropriate imo.

nicolas-chaulet:
You can put anything you want in get_input() or get_output() and kpconv
of=r Minkowski could have 2 different implementations that's fine. The
common point here is that xyz is the raw position of your input, so to me
looks like get_input is a good option. In the rest of the code it returns
untransformed input passed to the network that can contain references to
the raw point cloud via the origin_id key for example. Same goes for
get_ind, it could be embedded into the get_output no?

PR #263, opened 2020-05-09
update config

tchaton:


torch_points3d/applications/conf/kpconv/unet_4.yaml
```diff
@@ -4,6 +4,7 @@ define_constants:
     in_grid_size: 0.02
     in_feat: 64
     bn_momentum: 0.2
+    output_nc: 256
```
nicolas-chaulet, line 7: Are you sure you want that? It breaks the symmetry of the unet, and if you want a different size it would probably be better to add a fully connected at the end no?

torch_points3d/applications/conf/pointnet2/unet_3_ms.yaml
```diff
@@ -1,4 +1,6 @@
 conv_type: "DENSE"
+define_constants:
+    output_nc: 256
```
nicolas-chaulet, line 3: Again, it is a bit weird for an encoder decoder

torch_points3d/applications/conf/rsconv/unet_4.yaml
```diff
@@ -1,4 +1,6 @@
 conv_type: "DENSE"
+define_constants:
+    output_nc: 256
```
nicolas-chaulet, line 3: Same

torch_points3d/applications/kpconv.py
```diff
@@ -7,21 +7,16 @@
...
 CUR_FILE = os.path.realpath(__file__)
 DIR_PATH = os.path.dirname(os.path.realpath(__file__))
 PATH_TO_CONFIG = os.path.join(DIR_PATH, "conf/kpconv")
 
 
 def KPConv(
-    architecture: str = None,
-    input_nc: int = None,
-    num_layers: int = None,
-    in_grid_size: float = 0.02,
```
nicolas-chaulet, line 20: I had left those two arguments explicitely for documentation purposes. I will add them back!

torch_points3d/applications/kpconv.py
```diff
@@ -140,4 +160,7 @@ def forward(self, data):
 
         """
         self._set_input(data)
-        return super().forward(self.input, precomputed_down=self.pre_computed, precomputed_up=self.upsample)
+        data = super().forward(self.input, precomputed_down=self.pre_computed, precomputed_up=self.upsample)
+        if self.contains_output_nc:
```
nicolas-chaulet, line 164: I'll rename `contains_output_nc` to `has_mlp_head` if that's ok. I think it translates better the idea that it is not only the vanilla network but it also has a head

torch_points3d/applications/kpconv.py
```diff
@@ -140,4 +160,7 @@ def forward(self, data):
 
         """
         self._set_input(data)
-        return super().forward(self.input, precomputed_down=self.pre_computed, precomputed_up=self.upsample)
+        data = super().forward(self.input, precomputed_down=self.pre_computed, precomputed_up=self.upsample)
```
nicolas-chaulet, line 163: We probably want to change the encoder as well

test/test_api.py
```diff
@@ -67,16 +102,23 @@ def test_rsconv(self):
...
         )
         dataset = MockDataset(input_nc, num_points=1024)
         self.assertEqual(len(model._modules["down_modules"]), num_layers)
         self.assertEqual(len(model._modules["inner_modules"]), 2)
         self.assertEqual(len(model._modules["up_modules"]), num_layers + 1)
 
         try:
-            model.forward(dataset[0])
+            data_out = model.forward(dataset[0])
+            self.assertEqual(data_out.x.shape[1], output_nc)
```
nicolas-chaulet, line 121: Add tests for the encoder

torch_points3d/applications/kpconv.py
tchaton: I removed them as they weren't passed within kwargs nor given to the function.
I removed them but not from the doc, so user can still pass them but they will be directly placed within kwargs, as the resolve is done with it.

torch_points3d/applications/kpconv.py
tchaton: Fine

torch_points3d/applications/kpconv.py
tchaton: Yes

test/test_api.py
tchaton: Yes

torch_points3d/applications/kpconv.py
nicolas-chaulet: yes 👍

PR #265, opened 2020-05-10
Unwrapped Minkowski + get_input

humanpose1:
I still need to test the Unwrapped minkowski, the model is loading but I still need to see if it converges or not.
The unwrapped version will be very handy to add some other heads to the network.


test/test_registration_tracker.py
```diff
@@ -76,11 +77,12 @@ def __init__(self):
     def get_output(self):
         return self.feat[self.iter], self.feat_target[self.iter]
 
-    def get_xyz(self):
-        return self.xyz[self.iter], self.xyz_target[self.iter]
-
-    def get_ind(self):
-        return self.ind[self.iter], self.ind_target[self.iter], self.ind_size[self.iter]
+    def get_input(self):
```
nicolas-chaulet, line 80: Nice!!

PR #267, opened 2020-05-12
Modelnet

nicolas-chaulet:
I had to create a new version of ModelNet, the one on PYG is not very good for point clouds. People seem to use that one instead. Sampled meshes are very clean and regular.

### DIFF
--- a/conf/data/classification/modelnet.yaml
+++ b/conf/data/classification/modelnet.yaml
@@ -6,6 +6,8 @@ data:
     number: 10
     pre_transforms:
         - transform: NormalizeScale
+        - transform: GridSampling
+            lparams: [0.02]
     train_transforms:
         - transform: FixedPoints
             lparams: [2048]
--- a/torch_points3d/datasets/base_dataset.py
+++ b/torch_points3d/datasets/base_dataset.py
@@ -20,6 +20,24 @@ from torch_points3d.utils.colors import COLORS, colored_print
 log = logging.getLogger(__name__)
 
 
+def explode_transform(transforms):
+    """ Returns a flattened list of transform
+        Arguments:
+            transforms {[list | T.Compose]} -- Contains list of transform to be added
+
+        Returns:
+            [list] -- [List of transforms]
+        """
+    if transforms is not None:
+        if isinstance(transforms, Compose):
+            out = transforms.transforms
+        elif isinstance(transforms, list):
+            out = transforms
+        else:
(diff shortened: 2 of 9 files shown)


torch_points3d/datasets/classification/modelnet.py
```diff
@@ -1,12 +1,113 @@
...
+
+    @property
+    def processed_file_names(self):
+        return ["training_{}.pt".format(self.name), "test_{}.pt".format(self.name)]
+
+    def download(self):
+        path = download_url(self.url, self.root)
+        extract_zip(path, self.root)
+        os.unlink(path)
+        folder = osp.join(self.root, "modelnet40_normal_resampled")
```
humanpose1, line 74: modelnet10 or modelnet40 right ?

conf/data/classification/modelnet.yaml
```diff
@@ -6,6 +6,8 @@ data:
     number: 10
     pre_transforms:
         - transform: NormalizeScale
+        - transform: GridSampling
+            lparams: [0.02]
```
humanpose1, line 10: Is a voxel size of 0.02 adapted for ModelNet  ?

torch_points3d/datasets/classification/modelnet.py
nicolas-chaulet: The zip file is the same, the distinction happens when creating the dataset

conf/data/classification/modelnet.yaml
nicolas-chaulet: Yes, it is in normalized coordinates (happens after the `NormalizeScale`). I am putting a notebook together to show all this. And it runs on Colab :)
https://colab.research.google.com/github/nicolas-chaulet/torch-points3d/blob/modelnettuto/dashboard/ObjectClassificationRSConv.ipynb

nicolas-chaulet: image

nicolas-chaulet:
The dataset if of better quality (at least visually) and contains normal vectors, so less risk of missing something and better quality

torch_points3d/datasets/base_dataset.py
```diff
@@ -20,6 +20,25 @@
 log = logging.getLogger(__name__)
 
 
+def explode_transform(transforms):
```
tchaton, line 23: We might want this one to be recursive no ? as a compose can have nested compose no ?

PR #300, opened 2020-06-15
Application/minkowski

humanpose1:
Small PR to add Minkowski in the API :) . I have also modify GlobalBaseModule so that we don't need pos attribute.
the yaml and the format is slightly different from kpconv.
What do you think ?


torch_points3d/applications/conf/minkowski/encoder_4.yaml
```diff
@@ -0,0 +1,26 @@
+
+class: minkowski.MinkowskiFragment
```
nicolas-chaulet, line 2: Would you mind renaming this MinkowskiFragment to something a bit more explicit? It is not specific to fragments right? So maybe MinkowskiResnet or something like that

torch_points3d/applications/minkowski.py
```diff
@@ -0,0 +1,187 @@
...
+
+CUR_FILE = os.path.realpath(__file__)
+DIR_PATH = os.path.dirname(os.path.realpath(__file__))
+PATH_TO_CONFIG = os.path.join(DIR_PATH, "conf/minkowski")
+
+log = logging.getLogger(__name__)
+
+
+def Minkowski(
+    architecture: str = None,
```
nicolas-chaulet, line 25: Could we add an option to build it based on the already implemented models? `ResUnet16A` etc...
Something like `model_name` and it could not be specified together with architecture.

humanpose1: I can but the implemented models are not UnwrappedBasedModel. I can add a third option (like ```architecture="predefined_unet"```). I think it's easier

nicolas-chaulet: Don't worry, I forgot about that

PR #303, opened 2020-06-16
Fix various little bug for s3dis training

nicolas-chaulet:


torch_points3d/core/data_transform/grid_transform.py
```diff
@@ -159,6 +159,9 @@ def __call__(self, data):
             data = self._process(data)
         return data
 
+    def __repr__(self):
```
tchaton, line 162: Can you add the params ?

nicolas-chaulet:
There is no param to that one :)

PR #304, opened 2020-06-17
Panoptic

nicolas-chaulet:
Panoptic segmentation with PointGroup
TODO:

- [x] Add clustering
- [ ] Add instance mAP metrics
- [x] Add scoring network

EDIT:
I'll take the mAP calculation as a separate PR. There is already a lot in there...


torch_points3d/applications/conf/minkowski/encoder_2.yaml
```diff
@@ -0,0 +1,26 @@
+
```
humanpose1, line 1: why encoder_2 ?

tchaton: 2 layers.

torch_points3d/datasets/panoptic/scannet.py
```diff
@@ -0,0 +1,165 @@
...
+from torch_points3d.metrics.panoptic_tracker import PanopticTracker
+from torch_points3d.datasets.base_dataset import BaseDataset
+
+DIR = os.path.dirname(os.path.realpath(__file__))
+
+log = logging.getLogger(__name__)
+
+
+class ScannetPanoptic(Scannet):
+    TYPE2CLASS = {
```
tchaton, line 16: I wonder if we could not create a BaseScannet and not copy those every time

torch_points3d/datasets/panoptic/scannet.py
```diff
@@ -0,0 +1,165 @@
...
+        self.CLASS2TYPE = {self.TYPE2CLASS[t]: t for t in self.TYPE2CLASS}
+        self.NYU40ID2CLASS = {nyu40id: i for i, nyu40id in enumerate(list(self.NYU40IDS))}
+
+    def __getitem__(self, idx):
+        """
+        Data object contains:
+            pos - points
+            x - features
+        """
+        if not isinstance(idx, int):
```
tchaton, line 51: data = super(). __getitem__(idx)
self._set_extra_labels(data)

 should work right ?

torch_points3d/datasets/panoptic/scannet.py
```diff
@@ -0,0 +1,165 @@
...
+        data.vote_label = point_votes.float()
+        data.instance_labels = instance_labels
+        data.instance_mask = instance_labels != 0
+        data.num_instances = torch.tensor([num_instances])
+
+        # Remap labels
+        data = super()._remap_labels(data)
+        return data
+
+    def _remap_labels(self, data):
```
tchaton, line 111: Why ?

torch_points3d/datasets/panoptic/scannet.py
```diff
@@ -0,0 +1,165 @@
...
+            transform=self.train_transform,
+            version=dataset_opt.version,
+            use_instance_labels=use_instance_labels,
+            use_instance_bboxes=False,
+            donotcare_class_ids=donotcare_class_ids,
+            max_num_point=max_num_point,
+            is_test=is_test,
+        )
+
+        self.val_dataset = ScannetPanoptic(
```
tchaton, line 144: Can you please add test to ? So it is handled since the beginning.

torch_points3d/models/panoptic/pointgroup.py
```diff
@@ -0,0 +1,92 @@
...
+from torch_points3d.core.common_modules import Seq, MLP
+
+
+class PointGroupResults(NamedTuple):
+    semantic_logits: torch.Tensor
+    offset_logits: torch.Tensor
+    cluster_scores: torch.Tensor
+
+
+class PointGrouplabels(NamedTuple):
```
tchaton, line 16: Great !

torch_points3d/applications/conf/minkowski/encoder_2.yaml
nicolas-chaulet: @humanpose1 yes it stands for a small two layer deep encoder. I tried to pick reasonable layer sizes, what do you think?

humanpose1: But I see 4 layers in this encoder...

abhigoku10:
@nicolas-chaulet hi do we have panoptic for semanttic kitti / autonomous driving dataset which you r working on

nicolas-chaulet:
Hi, no, not yet. And we don't have any of the autonomous driving dataset integrated yet. This is certainly of interest so feel free to contribute by adding one! Happy to provide guidance.

torch_points3d/applications/conf/minkowski/encoder_2.yaml
nicolas-chaulet: Good catch... Removing for now.

torch_points3d/datasets/panoptic/scannet.py
nicolas-chaulet: Unused. I removed it

torch_points3d/datasets/panoptic/scannet.py
nicolas-chaulet: Yes correct, keeping the check on idx though.

torch_points3d/datasets/panoptic/scannet.py
nicolas-chaulet: This is called at the end of the __init__ method of the parent class and I don't want it to be. Labels are computed on the fly after the transformation has been applied

torch_points3d/datasets/panoptic/scannet.py
nicolas-chaulet: I'll test the dataset yes.

torch_points3d/models/panoptic/pointgroup.py
```diff
@@ -0,0 +1,177 @@
...
+        self.labels = data.y.to(device)
+        all_labels = {l: data[l].to(device) for l in self.__REQUIRED_LABELS__}
+        self.labels = PanopticLabels(**all_labels)
+
+    def forward(self, epoch=-1, **kwargs):
+        # Backbone
+        backbone_features = self.Backbone(self.input).x
+
+        # Semantic and offset heads
+        semantic_logits = self.Semantic(backbone_features)
```
tchaton, line 47: I am not sure it is how the model should be. I thought the model predicted N + 3 features directly out of the backbone. Here backbone_features is used for 2 tasks at the same time, it might lose some expressivity.

test/mockdatasets.py
```diff
@@ -77,6 +85,10 @@ def __getitem__(self, index):
     def class_to_segments(self):
         return {"class1": [0, 1, 2, 3, 4, 5], "class2": [6, 7, 8, 9]}
 
+    @property
```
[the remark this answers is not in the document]
nicolas-chaulet: This are classes that should be ignored for box or instance segmentation but you still want them in the semantic segmentation

torch_points3d/applications/minkowski.py
```diff
@@ -111,11 +105,14 @@ def _set_input(self, data):
         """
         coords = torch.cat([data.batch.unsqueeze(-1).int(), data.coords.int()], -1)
         self.input = ME.SparseTensor(data.x, coords=coords).to(self.device)
-        self.xyz = data.pos.to(self.device)
+        if data.pos is not None:
```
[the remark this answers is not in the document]
nicolas-chaulet: Yes, xyz is the raw position, not the quantized one

torch_points3d/datasets/panoptic/scannet.py
```diff
@@ -0,0 +1,156 @@
...
+log = logging.getLogger(__name__)
+
+
+class ScannetPanoptic(Scannet):
+    NYU40IDS = np.array([3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 14, 16, 24, 28, 33, 34, 36, 39])
+    NUM_MAX_OBJECTS = 64
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.STUFFCLASSES = torch.tensor([i for i in self.VALID_CLASS_IDS if i not in self.NYU40IDS])
```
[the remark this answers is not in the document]
nicolas-chaulet: I have seen that terminology used in a few image papers. Those are classes that are ignored for certain tasks (in that case instance segmentation)

torch_points3d/models/panoptic/pointgroup.py
nicolas-chaulet: I checked the original repo and it is done that way, the same set of features goes into two linear networks.

PR #307, opened 2020-06-23
Transform/rwmask

humanpose1:
A small PR to add random walk Dropout. the transform build a graph using radius nn search and drop randomly vertices on a graph.


humanpose1:
rw_001
rw_0001
rw_001_iter_30000
rw_0001_iter_100000
rw_005
rw_low_prob

Here are some examples in KITTI Odometry dataset using different params

PR #310, opened 2020-06-26
Amending behaviour of grid sampling in the case of sparse coordinates

nicolas-chaulet:
Implements what we discussed. @humanpose1 I believe I have handled the registration bit correctly but please double check!

### DIFF
--- a/torch_points3d/models/registration/minkowski.py
+++ b/torch_points3d/models/registration/minkowski.py
@@ -52,9 +52,9 @@ class BaseMinkowski(FragmentBaseModel):
         self.input = ME.SparseTensor(data.x, coords=coords).to(device)
         self.xyz = data.pos.to(device)
         if hasattr(data, "pos_target"):
-            coords_target = torch.cat([data.batch_target.unsqueeze(-1).int(), data.pos_target.int()], -1)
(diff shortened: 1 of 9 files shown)


torch_points3d/models/registration/minkowski.py
```diff
@@ -48,9 +48,9 @@ def __init__(self, option, model_type, dataset, modules):
             self.FC_layer = torch.nn.Identity()
 
     def set_input(self, data, device):
-        coords = torch.cat([data.batch.unsqueeze(-1).int(), data.pos.int()], -1)
+        coords = torch.cat([data.batch.unsqueeze(-1).int(), data.coords.int()], -1)
         self.input = ME.SparseTensor(data.x, coords=coords).to(device)
-        self.xyz = torch.stack((data.pos_x, data.pos_y, data.pos_z), 0).T.to(device)
+        self.xyz = data.pos.to(device)
         if hasattr(data, "pos_target"):
             coords_target = torch.cat([data.batch_target.unsqueeze(-1).int(), data.pos_target.int()], -1)
```
humanpose1, line 55: You forget to change pos_target into coords_target, the input of the network is
PairBatch(pos, x, batch, coords, pos_target, x_target, batch_target, coords_target)

nicolas-chaulet: Well spotted, thanks. It should be good now

PR #315, opened 2020-07-01
added fix for kernel image fail

loicland:
Documented the crash that can happen when compiling with different GPUs, + a fix and a link to a list of architecture.

This is not an issue specific to this repo in particular, but torch-points-kernels does fail when using Turing -> Volta, and it is not well documented elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,14 @@ rm -rf ~/.cache/pip
 poetry install
 ```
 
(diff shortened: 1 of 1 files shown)


nicolas-chaulet:
Thanks @loicland Moving that to the troubleshooting section :)

PR #322, opened 2020-07-03
Train Dockerfile

CCInc:
Here's the base train dockerfile... I've confirmed it to work for `pip install -r` but not `pip install .`, let me know what your ideas are.

### DIFF
--- a/docker/Dockerfile.train
+++ b/docker/Dockerfile.train
@@ -17,18 +17,13 @@ RUN python3 -m pip install -U pip \
     && pip3 install pycuda \
     && rm -rf /root/.cache
 
-#COPY requirements.txt requirements.txt
-COPY pyproject.toml pyproject.toml
+ENV WORKDIR=/tp3d
+WORKDIR $WORKDIR
 
-#RUN FORCE_CUDA=1 TORCH_CUDA_ARCH_LIST="3.5 5.2 6.0 6.1 7.0+PTX" pip3 install -r requirements.txt \
-#    && rm -rf /root/.cache
+COPY pyproject.toml pyproject.toml
+COPY torch_points3d/__init__.py torch_points3d/__init__.py
+COPY README.md README.md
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.train
```diff
@@ -0,0 +1,34 @@
...
+
+RUN python3 -m pip install -U pip \
+    && python3 -m pip install -U pip \
+    && pip3 install setuptools>=41.0.0 \
+    && pip3 install torch==1.5.0+cu101 torchvision==0.6.0+cu101 -f https://download.pytorch.org/whl/torch_stable.html \
+    && pip3 install git+git://github.com/StanfordVL/MinkowskiEngine.git#v0.4.3 \
+    && pip3 install pycuda \
+    && rm -rf /root/.cache
+
+#COPY requirements.txt requirements.txt
```
nicolas-chaulet, line 20: Create a workdir here, otherwise it is at the root location, that's why the pip install . fails.
```
WORDIR /tp3d
```
Then everything should work

docker/Dockerfile.train
```diff
@@ -0,0 +1,34 @@
...
+
+#RUN FORCE_CUDA=1 TORCH_CUDA_ARCH_LIST="3.5 5.2 6.0 6.1 7.0+PTX" pip3 install -r requirements.txt \
+#    && rm -rf /root/.cache
+
+RUN FORCE_CUDA=1 TORCH_CUDA_ARCH_LIST="3.5 5.2 6.0 6.1 7.0+PTX" pip3 install . && rm -rf /root/.cache
+
+# we need to specify TORCH_CUDA_ARCH_LIST to compile with non-gpu docker machines
+# this list was taken from pytorch's list of deafult architectures (https://github.com/pytorch/pytorch/blob/master/docker/pytorch/Dockerfile#L28)
+
+ENV WORKDIR=/dpb
```
nicolas-chaulet, line 31: Remove

docker/Dockerfile.train
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR $WORKDIR
```
nicolas-chaulet, line 33: Done before you run `pip install .`

CCInc:
Yep, that was it! I was able to build and run the unit tests with this update.

PR #337, opened 2020-07-11
Pretrained registry

tchaton:
Draft for the registry. I have added a used_save_properties. During creation, the property are being wrapped and the output cached into this dictionary. This dict should be saved and reload from checkpoint, so we can mock dataset.

@nicolas-chaulet Can you check if you like this solution, I will finish it. If not, we will think about something else.

```
model = PretainedRegistry.from_pretrained(
            "pointnet2_largemsg", "s3dis", fold=1, download=True)

print(model.wandb)
```

```
(tp3d) ➜  torch-points3d git:(pretrained_registry) python test/test_api.py TestPretainedRegistry
WARNING: skipping download of existing file /Users/tcn02/projects/perso/torch-points3d-templates/torch-points3d/torch_points3d/applications/weights/pointnet2_largemsg_s3dis_1.pt
=================================================== WANDB URLS ===================================================================
WEIGHT_URL: https://api.wandb.ai/files/loicland/benchmark-torch-points-3d-s3dis/1e1p0csk/pointnet2_largemsg.pt
LOG_URL: https://app.wandb.ai/loicland/benchmark-torch-points-3d-s3dis/runs/1e1p0csk/logs
CHART_URL: https://app.wandb.ai/loicland/benchmark-torch-points-3d-s3dis/runs/1e1p0csk
OVERVIEW_URL: https://app.wandb.ai/loicland/benchmark-torch-points-3d-s3dis/runs/1e1p0csk/overview
HYDRA_CONFIG_URL: https://app.wandb.ai/loicland/benchmark-torch-points-3d-s3dis/runs/1e1p0csk/files/hydra-config.yaml
OVERRIDES_URL: https://app.wandb.ai/loicland/benchmark-torch-points-3d-s3dis/runs/1e1p0csk/files/overrides.yaml
=================================================================================================================================

.
----------------------------------------------------------------------
Ran 1 test in 0.150s
```


torch_points3d/applications/pretrained_api.py
```diff
@@ -0,0 +1,114 @@
...
+
+            checkpoint: ModelCheckpoint = ModelCheckpoint(
+                CHECKPOINT_DIR,
+                checkpoint_model_name,
+                weight_name if weight_name is not None else "latest",
+                resume=False,
+            )
+            if mock_dataset:
+                dataset = instantiate_mock_dataset(
+                    checkpoint.data_config, {"feature_dimension": 4, "num_classes": 13})
```
humanpose1, line 98: I don't understand whether this class is specific for s3dir or it is generic (feature dimension of 4 depends on the model)

test/test_api.py
```diff
@@ -9,15 +13,42 @@
...
-from test.mockdatasets import MockDatasetGeometric, MockDataset
-from torch_points3d.core.data_transform import GridSampling3D
-from torch_points3d.datasets.object_detection.scannet import ScannetDataset
-
 seed = 0
 torch.manual_seed(seed)
 device = "cpu"
 
 
+class TestPretainedRegistry(unittest.TestCase):
```
humanpose1, line 21: Maybe add some tests for other datasets

torch_points3d/applications/pretrained_api.py
```diff
@@ -0,0 +1,114 @@
...
+        dataset_name = dataset_name.lower()
+        fold = str(fold) if fold is not None else fold
+
+        model = PretainedRegistry.__dict__.get(model_name)
+        if model:
+            dataset = model.get(dataset_name)
+            if dataset:
+                if isinstance(dataset, dict):
+                    if fold is None:
+                        raise Exception(
```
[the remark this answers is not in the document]
tchaton: It won't use it if dataset isn't a dict

torch_points3d/applications/pretrained_api.py
tchaton: Yes ahah it is still a draft. I am just trying to make it work for now :)

test/test_api.py
tchaton: When more pretrained models will be available, I will add them. Would be great to get some of yours.

test/test_api.py
```diff
@@ -63,7 +95,8 @@ def test_kpconv(self):
             num_layers=num_layers,
             config=None,
         )
-        dataset = MockDatasetGeometric(input_nc + 1, transform=GridSampling3D(0.01), num_points=128)
+        dataset = MockDatasetGeometric(
```
nicolas-chaulet, line 98: Formating is wrong, could you make sure that you ahve pre-commit properly activated?

torch_points3d/applications/pretrained_api.py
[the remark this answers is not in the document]
nicolas-chaulet: I agree with @humanpose1, I think fold should be part of the dataset name "s3dis-fold1" for example. This will remove some complexity and hard coding to s3dis

torch_points3d/applications/pretrained_api.py
```diff
@@ -0,0 +1,114 @@
...
+    if not os.path.exists(out_file):
+        urllib.request.urlretrieve(url, out_file)
+    else:
+        log.warning("WARNING: skipping download of existing file " + out_file)
+
+
+class PretainedRegistry(object):
+
+    pointnet2_largemsg = {
+        "s3dis": {
```
nicolas-chaulet, line 32: Flatten this to "s3dis-fold1" etc... See comment below

torch_points3d/datasets/base_dataset.py
```diff
@@ -36,10 +37,23 @@ def explode_transform(transforms):
         elif isinstance(transforms, list):
             out = copy.deepcopy(transforms)
         else:
-            raise Exception("transforms should be provided either within a list or a Compose")
+            raise Exception(
```
nicolas-chaulet, line 40: Please verify formatting, your pre-commit is probably not setup

torch_points3d/utils/wandb_utils.py
```diff
@@ -3,16 +3,49 @@
...
+            msg += "{}: {}\n".format(k.upper(), v)
+        msg += "=================================================================================================================================\n"
+        return msg
+
+
 class Wandb:
     IS_ACTIVE = False
 
-    @staticmethod
+    @ staticmethod
```
nicolas-chaulet, line 37: There shouldn't be a space here right?

torch_points3d/utils/wandb_utils.py
```diff
@@ -3,16 +3,49 @@
...
+        model.wandb = wandb_urls
+
+    @ staticmethod
     def _set_to_wandb_args(wandb_args, cfg, name):
         var = getattr(cfg.wandb, name, None)
         if var:
             wandb_args[name] = var
 
-    @staticmethod
+    @ staticmethod
```
nicolas-chaulet, line 48: Same, why space?

torch_points3d/utils/wandb_utils.py
```diff
@@ -41,8 +74,10 @@ def launch(cfg, launch: bool):
             Wandb._set_to_wandb_args(wandb_args, cfg, "config")
 
             try:
-                commit_sha = subprocess.check_output(["git", "rev-parse", "HEAD"]).decode("ascii").strip()
-                gitdiff = subprocess.check_output(["git", "diff", "--", "':!notebooks'"]).decode()
+                commit_sha = subprocess.check_output(
```
nicolas-chaulet, line 77: Bad formatting

torch_points3d/datasets/dataset_factory.py
```diff
@@ -44,3 +45,12 @@ def instantiate_dataset(dataset_config) -> BaseDataset:
     dataset_cls = get_dataset_class(dataset_config)
     dataset = dataset_cls(dataset_config)
     return dataset
+
+
+def instantiate_mock_dataset(dataset_config, used_properties) -> PretrainedMockDataset:
+    """Import the module "data/[module].py".
```
nicolas-chaulet, line 51: Please revise docstring

torch_points3d/datasets/base_dataset.py
```diff
@@ -36,10 +37,23 @@ def explode_transform(transforms):
...
+                "transforms should be provided either within a list or a Compose")
     return out
 
 
+def save_used_properties(func):
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if self.used_properties.get(func.__name__) is None:
+            result = func(self, *args, **kwargs)
+            self.used_properties[func.__name__] = result
```
nicolas-chaulet, line 50: This is really dangerous. Imagine the property changes compared to the value that you store here? I think that  we should never cache the results of properties or functions. If you just return result here it will do what you want without the risk right?

torch_points3d/applications/pretrained_api.py
nicolas-chaulet: If I understand correctly here you are using that simply to mock a dataset so that the model init method can access deature_dimension and num_classes right? If that's the case, then I would recommend creating some dummy classes that just contain this info and nothing else. A NamedTuple would do for example. Basically I don't think we should use the BaseDataset here (unless we really have to). It will create problems later down the line for sure.

test/test_api.py
```diff
@@ -21,30 +21,11 @@
...
-
-        for i in range(1, 7):
-            _ = PretainedRegistry.from_pretrained(
-                "pointnet2_largemsg", "s3dis", fold=i, download=False)
-
-        model = PretainedRegistry.from_pretrained(
-            "pointnet2_largemsg", "s3dis", fold=1, download=True)
+        model = PretainedRegistry.from_pretrained("pointnet2_largemsg-s3dis-1", download=True)
+
+        print(model.wandb)
```
nicolas-chaulet, line 26: Please remove those prints. We don't need them

torch_points3d/applications/pretrained_api.py
```diff
@@ -0,0 +1,93 @@
...
+        if not os.path.exists(os.path.dirname(out_file)):
+            os.makedirs(os.path.dirname(out_file))
+        urllib.request.urlretrieve(url, out_file)
+    else:
+        log.warning("WARNING: skipping download of existing file " + out_file)
+
+
+class PretainedRegistry(object):
+
+    MODELS = {
```
humanpose1, line 33: I don't know if it is handy to have this dictionnary hard coded in the code. Why not a csv or something like ?

torch_points3d/applications/pretrained_api.py
```diff
@@ -0,0 +1,93 @@
...
+        "pointnet2_largemsg-s3dis-2": {"feature_dimension": 4, "num_classes": 13},
+        "pointnet2_largemsg-s3dis-3": {"feature_dimension": 4, "num_classes": 13},
+        "pointnet2_largemsg-s3dis-4": {"feature_dimension": 4, "num_classes": 13},
+        "pointnet2_largemsg-s3dis-5": {"feature_dimension": 4, "num_classes": 13},
+        "pointnet2_largemsg-s3dis-6": {"feature_dimension": 4, "num_classes": 13},
+        "pointgroup-scannet": {},
+    }
+
+    @staticmethod
+    def from_pretrained(model_tag, download=True, out_file=None, weight_name="latest", mock_dataset=True):
```
humanpose1, line 54: There are at least two cases: Either, I want to download the pretrained model on wandb. Or I want to use a local .pt for my model. Do you have the intention to implement both cases ?

humanpose1: Maybe you can add an other parameter such as `path_model`, if I want to use a local pretrained model.

torch_points3d/applications/pretrained_api.py
nicolas-chaulet: Yes, let's do that in a second step once we understand use cases better. Good shout!

torch_points3d/applications/pretrained_api.py
nicolas-chaulet: Good point as well, I was thinking about using something like that for PointGroup for example.

PR #264, opened 2020-05-09
PointAugment: an Auto-Augmentation Framework for Point Cloud Classification

tchaton:
https://arxiv.org/pdf/2002.10876.pdf

Please, verify the code.

I have still a pb with KPConv
```
    allow_unreachable=True)  # allow_unreachable flag
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.LongTensor [521, 38, 3]] is at version 2; expected version 1 instead. Hint: the backtrace further above shows the operation that failed to compute its gradient. The variable in question was changed in there or anywhere later. Good luck!
```

PR #339, opened 2020-07-14
fix wrong reference PointAugmentation arxiv.org

roma-goodok:


nicolas-chaulet:
Hey, thanks for that! I'll remove it entirely we have a pending PR with a tentative implementation here #264 but not sure it actually provides the results claimed. So at the moment we actually don't have PointAugment. If you'd like to add it or contribute feel free to ping us!

nicolas-chaulet:
Sorry, missed the fact that you already spotted the PR! Merging now

Issue #328, opened 2020-07-06
LR secheduler resume is broken for update on batch number


nicolas-chaulet:
When resuming a training, the counter of number of batches in base_model is set to 0 (as opposed to the number of processed batches). As a consequence the schedulers that use this property are broken when resuming a training.

PR #341, opened 2020-07-15
resolve scheduler bug

tchaton:
https://github.com/nicolas-chaulet/torch-points3d/issues/328


torch_points3d/metrics/model_checkpoint.py
```diff
@@ -287,5 +290,11 @@ def save_best_models_under_current_metrics(
                     current_stat["best_{}".format(metric_name)] = metric_value
                     models_to_save["best_{}".format(metric_name)] = state_dict
 
+        kwargs["model_props"] = {
+            "_num_epochs": model._num_epochs,  # type: ignore
```
nicolas-chaulet, line 298: Could you remove the _ from their name since you access those directy? Also the base model defines some unused members
```
 self._latest_metrics = None
        self._latest_stage = None
        self._latest_epoch = None
```
Could you please delete them while you're at it?

PR #342, opened 2020-07-17
Added feature normalization transform

CCInc:
Let me know if this transform would be useful for the repository :)

It just performs a minmax normalization on an arbitrary 1D feature. I suppose it could be extended to a differently shaped feature, but I'm not sure how normalization works in that case.


torch_points3d/core/data_transform/feature_augment.py
```diff
@@ -24,6 +24,31 @@ def __call__(self, data):
     def __repr__(self):
         return "{}({})".format(self.__class__.__name__, self._normalize)
 
+class NormalizeFeature(object):
```
tchaton, line 27: Dear @CCInc, I think it would be better to avoid this transform.

When scaling features, it is usually done on the entire dataset. Here, it is scaling on a batch and it tends to hurt performance.
Might be more interesting to do a dataset transform using Sklearn scalers.

Best regards,
Thomas Chaton

CCInc: @tchaton Yes ideally it would be applied on pre_transform or pre_collate_transform, not on the batch transforms :)

nicolas-chaulet: Yes makes sense. Since it is specific to pre collate, could you please create a new python file (maybe `precollate.py` and move it there? Like that it will avoid any confusion. Also could you please add a test?

CCInc: @nicolas-chaulet I made the changes, let me know if it's okay

nicolas-chaulet: Great thanks!

PR #346, opened 2020-07-21
Panoptic metrics

nicolas-chaulet:
Metrics for panoptic segmentation. Scannet is supported at the moment, but any dataset complying with the interface will do as well.


torch_points3d/metrics/panoptic_tracker.py
```diff
@@ -1,25 +1,291 @@
...
+            scan_id = pred.scan_id
+            gts = allgts.get(scan_id, [])
+            if len(gts) == 0:
+                fp[p] = 1
+                continue
+
+            # Find best macth in ground truth
+            ioumax = -np.inf
+            best_match = -1
+            for i, gt in enumerate(gts):
```
tchaton, line 69: Could you make a small function in _Instance. Find_best_match ?

PR #366, opened 2020-07-28
add PPNet (PosPool)

zeliu98:


conf/training/ppnet.yaml
```diff
@@ -0,0 +1,41 @@
...
+            class: SGD
+            params:
+                momentum: 0.98
+                lr: ${training.optim.base_lr} # The path is cut from training
+                weight_decay: 1e-3
+        lr_scheduler: ${lr_scheduler}
+        bn_scheduler:
+            bn_policy: "step_decay"
+            params:
+                bn_momentum: 0.99
```
nicolas-chaulet, line 21: This should be 0.01

torch_points3d/models/segmentation/ppnet.py
```diff
@@ -0,0 +1,155 @@
...
+        self.visual_names = ["data_visual"]
+        self.init_weights()
+
+    def set_input(self, data, device):
+        """Unpack input data from the dataloader and perform necessary pre-processing steps.
+        Parameters:
+            input: a dictionary that contains the data itself and its metadata information.
+        """
+        data = data.to(device)
+        data.x = add_ones(data.pos, data.x, True)
```
humanpose1, line 83: maybe you don't need to add_ones here because you can add [ones](https://github.com/nicolas-chaulet/torch-points3d/blob/fb6ef0d2ea3947bc60588cd8bc01d3c333c710f4/torch_points3d/core/data_transform/features.py#L292) in the data augmentation. You also need [that](https://github.com/nicolas-chaulet/torch-points3d/blob/fb6ef0d2ea3947bc60588cd8bc01d3c333c710f4/torch_points3d/core/data_transform/features.py#L162).

torch_points3d/modules/PPNet/ops.py
```diff
@@ -0,0 +1,175 @@
+import torch
+import torch.nn as nn
+from torch_points3d.core.common_modules.base_modules import FastBatchNorm1d
+
+
+def add_ones(query_points, x, add_one):
```
humanpose1, line 6: same as above. We can add ones in the data preprocessing step using the transforms

conf/training/ppnet.yaml
zeliu98: OK. I will change it and test it again.

torch_points3d/models/segmentation/ppnet.py
zeliu98: > Great PR ! Thanks a lot for your contributions ! I wrote some little comments below. I have few questions.
> -Have you tested PPNet on S3DIS or Shapenet to see if you could reproduce the results of your paper ?
I have tested it using the default data(train\val\test split) and training setting on S3DIS. The result is 78.3 mIoU(val) and 62.8 mIoU(test). This is similar with the kpconv benchmark results(https://app.wandb.ai/nicolas/s3dis-benchmark/runs/1lqzkigl/logs?workspace=user-nicolas). Considering the results reported in KPCONV paper is similar to ours, the reproduced results is reasonable. Another difference is that we use a voting mechanism in our code but this repo may not(I didn't find it). Besides, some data augmentation parameters may be different.

> maybe you don't need to add_ones here because you can add [ones](https://github.com/nicolas-chaulet/torch-points3d/blob/fb6ef0d2ea3947bc60588cd8bc01d3c333c710f4/torch_points3d/core/data_transform/features.py#L292) in the data augmentation. You also need [that](https://github.com/nicolas-chaulet/torch-points3d/blob/fb6ef0d2ea3947bc60588cd8bc01d3c333c710f4/torch_points3d/core/data_transform/features.py#L162).
OK, thanks for your suggestion. I will remove it.

zeliu98: > maybe you don't need to add_ones here because you can add [ones](https://github.com/nicolas-chaulet/torch-points3d/blob/fb6ef0d2ea3947bc60588cd8bc01d3c333c710f4/torch_points3d/core/data_transform/features.py#L292) in the data augmentation. You also need [that](https://github.com/nicolas-chaulet/torch-points3d/blob/fb6ef0d2ea3947bc60588cd8bc01d3c333c710f4/torch_points3d/core/data_transform/features.py#L162).

Hi, can you give me an example about how to use it? 
Is it just adding this lines in the data config like this?

"""
    - transform: XYZFeature
      params:
        add_x: False
        add_y: False
        add_z: True
    - transform: AddOnes
    - transform: AddFeatsByKeys
      params:
        list_add_to_x: [True, True]
        feat_names: [rgb, pos_z]
        delete_feats: [True, True]
"""

humanpose1:
```yaml
- transform: AddOnes
- transform: AddFeatsByKeys
params:
list_add_to_x: [True, True, True]
feat_names: [rgb, pos_z, ones]
delete_feats: [True, True, True]
```
AddOnes creates `ones` atribute in `Data`
Then you add the variables you want with `AddFeatsByKeys`.

zeliu98: got it, thanks a lot!

humanpose1: you're welcome :)

torch_points3d/modules/PPNet/ops.py
```diff
@@ -0,0 +1,175 @@
...
+    if add_one:
+        ones = torch.ones(query_points.shape[0], dtype=torch.float).unsqueeze(-1).to(query_points.device)
+        if x is not None:
+            x = torch.cat([ones.to(x.dtype), x], dim=-1)
+        else:
+            x = ones
+    return x
+
+
+def gather(x, idx, method=2):
```
tchaton, line 16: This function and the one from KPConv: https://github.com/nicolas-chaulet/torch-points3d/blob/c1e494ee205171b566f6f12000adada055d372d3/torch_points3d/modules/KPConv/convolution_ops.py#L8 should move in the torchpoints3d/core to reduce duplicated code.

torch_points3d/models/segmentation/ppnet.py
zeliu98: I have already changed them.

conf/training/ppnet.yaml
zeliu98: Hi @nicolas-chaulet. I found a little performance drop by changing it to 0.01.  
I wonder whether the `bn_momentum` here means the `momentum` in pytorch `nn.BatchNorm` or the  `momentum` in tensorflow `tf.layers.batch_normalization` since the `momentum` in pytorch is equivalent to a value of `1-momentum` in tensorflow. I'm a little confused.
Thanks a lot!

nicolas-chaulet: Here it means the momenum in PyTorch: https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html#torch.nn.BatchNorm1d. So setting it 0.99 means that you keep 99% of the old value and 2% of the new. If it works better then be it but it is quite weird

torch_points3d/modules/PPNet/ops.py
zeliu98: Hi @tchaton.
Should I add a new file like `gathering.py` in torchpoints3d/core and change the use both in KPConv and my code?

conf/training/ppnet.yaml
zeliu98: got it. I will keep it and maybe do more experiments later to find why.

torch_points3d/modules/PPNet/ops.py
zeliu98: I have already changed them

torch_points3d/modules/PPNet/ops.py
nicolas-chaulet: Yes sounds good

chaitjo:
For PPNet on S3DIS, is it recommended to use the data configuration file `conf/data/segmentation/s3disfused-addones.yaml`?

PR #370, opened 2020-07-28
Viz

tchaton:


scripts/visualizations/scannet_panoptic.py
```diff
@@ -0,0 +1,120 @@
...
+        if not existing_colors:
+            return color
+        best_distance = min([color_distance(color, c) for c in existing_colors])
+        if not max_distance or best_distance > max_distance:
+            max_distance = best_distance
+            best_color = color
+    return best_color
+
+
+def torch2o3d(data, color=[1, 0, 0]):
```
humanpose1, line 64: this function will be useful in a lot of different context. Can you put it in a util file a something like that ?

tchaton: Yes, will do.

Issue #368, opened 2020-07-28
How to change n_kernel_points in kpconv ?


sidml:
Hi.
I was wondering if it's possible to change n_kernel_points in kpconv.
I tried passing n_kernel_points in kpconv as a parameter while defining the model but it doesn't work. 
I also tried adding a new field named n_kernel_points in kpconv in the yaml file.
Any help is appreciated. Thanks.

tchaton:
Dear @sidml, yes it is possible. 

Not sure we connected it properly from configuration file to the KPConvLayer class creation.

Here is the `n_kernel_points ` params: 
https://github.com/nicolas-chaulet/torch-points3d/blob/94454627a3d5243b586fe1452f8071a9f2e1caa1/torch_points3d/modules/KPConv/kernels.py#L42

Would you mind connecting it to the configuration and opening a PR ?

Best regards,
Thomas Chaton.

PR #373, opened 2020-07-30
allow to set kpconv params from yaml file.

sidml:
This commit should allow user to configure KPConvLayer and KPConvDeformableLayer by modifying the yaml file.
https://github.com/nicolas-chaulet/torch-points3d/issues/368

### DIFF
--- a/torch_points3d/applications/conf/kpconv/unet_4.yaml
+++ b/torch_points3d/applications/conf/kpconv/unet_4.yaml
@@ -1,5 +1,7 @@
 class: kpconv.KPConvPaper
 conv_type: "PARTIAL_DENSE"
+KPConvLayer:
+    n_kernel_points: 15
 define_constants:
     in_grid_size: 0.02
     in_feat: 64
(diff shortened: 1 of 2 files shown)


torch_points3d/applications/conf/kpconv/unet_4.yaml
```diff
@@ -1,5 +1,7 @@
 class: kpconv.KPConvPaper
 conv_type: "PARTIAL_DENSE"
+KPConvLayer:
```
tchaton, line 3: Thanks for contributing. I need to check, but I think there is another solution to make this work. The unet should be model agnostic.

tchaton: Let me check on my side. @nicolas-chaulet Do you confirm there is a better way ?

nicolas-chaulet: Yes, there is, let me create a PR for that, it will just be faster

nicolas-chaulet:
See #378

PR #375, opened 2020-07-30
added vispy vis and semantickitti utils

MukundVarmaT:
Details of pull request
- Added semantic kitti dataset which includes a Lidar class for projection and reprojection of points, labels, a data loader and other dataset details.
- Added vispy support for visualising such large lidar scans. Since the point cloud has over 122k points, rendering at high FPS while enabling panning and other user controls is quite impossible on PyQt, Mayavi or other cpu based visualisation tools. Vispy is a GPU based visualisation tool. Here is an example vis of the same generated using the following code:
```
from torch_points3d.visualization import vispy_vis
from torch_points3d.datasets.segmentation import semantic_kitti
import os
from glob import glob
import time
from skimage.transform import resize
import numpy as np
import cv2

class VisSemKitti(vispy_vis.Vispy):
    def __init__(self, use_default=True, mode='sem_kitti'):
        super().__init__(use_default=use_default, mode=mode)
        self.sequence_dir = "/home/sneezygiraffe/Herbie/segmentation/semantic_kitti/sequences/00"
        self.velodyne_files = sorted(glob(os.path.join(self.sequence_dir, "velodyne", '*.bin')))
        self.label_files = sorted(glob(os.path.join(self.sequence_dir, "labels", '*.label')))
        self.file_counter = 0
        self.semantickittiscan = semantic_kitti.LidarScan(64, 2048, 3, -25)
        self.semantickittiscan.populate_colormap()

    def update(self, event):
        points, remissions = self.semantickittiscan.open_scan(self.velodyne_files[self.file_counter])
        start = time.time()
        _, _, _, _, _, proj_index, proj_mask, proj_x, proj_y = self.semantickittiscan.project_scan(points, remissions)
        unproj_sem_label, _ = self.semantickittiscan.open_label(self.label_files[self.file_counter])
        proj_sem_label = self.semantickittiscan.project_sem_label(proj_index, proj_mask, unproj_sem_label)
        color_proj = self.semantickittiscan.colorise_projection(proj_sem_label, proj_mask, reduced=True)
        color_unproj = self.semantickittiscan.colorise_points(unproj_sem_label, reduced=True)
        self.pc_vis.set_data(points, face_color=color_unproj, edge_color=color_unproj, size=2)
        color_proj = resize(color_proj, (64, 1920))
        cv2.putText(color_proj, "FPS : "+str(1//(time.time()-start)), (30,20), cv2.FONT_HERSHEY_SIMPLEX, 0.7, (255,0,0), 3)
        self.img_vis.set_data(color_proj)
        self.file_counter +=1

vis = VisSemKitti()
vis.run()
```
and the output looks like this : 

ezgif com-optimize (1)

I am getting a constant 15 FPS which is quite good. PS. i moved around the screen a lot just to show how quick it is while rendering and there are no FPS drops.

I was quite unsure about how to generalise this and hence I though we can add some default layouts for certain visualisations which require this tool (meaning where we need to plot more points and at high FPS). For now I have added a layout for semantic kitti where the top box is for the unprojected point cloud, while the bottom is for the projected image. Since there is a lot of variability in what a user might plot i thought it will be a better idea to leave the update function for the user to write based on requirement.

I will add some models and training scripts for semantic kitti next. I thought of implementing this [paper](https://arxiv.org/abs/2003.03653) which is the current state of the art for segmentation in the kitti dataset.


torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+        proj_color_labels = np.zeros((self.H, self.W, 3), dtype=np.float)
+        if reduced:
+            proj_label = np.vectorize(learning_map.get)(proj_label)
+            proj_color_labels[proj_mask] = self.reduced_colormap[proj_label[proj_mask]]
+        else:
+            proj_color_labels[proj_mask] = self.colormap[proj_label[proj_mask]]
+        return proj_color_labels
+
+
+class SemanticKitti(Dataset):
```
tchaton, line 360: Could you reformat this code as in ShapeNet.
Kitti will be using `from torch_geometric.data import Dataset`. You can find its implementation there: https://pytorch-geometric.readthedocs.io/en/latest/modules/data.html#torch_geometric.data.Dataset

KittiDataset will be using Kitti dataset for train, val and test as in ShapeNetDataset with their own pre_transform, transform.

Also, could you add a conf/data/segmentation/kitti.yaml file with any hyper-parameters you might think about. Take examples on the other ones.

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+  "test"  : [11,12,13,14,15,16,17,18,19,20,21]
+}
+
+sensor_config = {
+    "name": "HDL64",
+    "type": "spherical",
+    "fov_up": 3,
+    "fov_down": -25
+}
+img_prop = {
```
tchaton, line 214: Are those values fixed ? If not, they should be within /conf/data/segmentation/kitti.yaml file as hyper-parameters of the dataset.

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
+"""
+Dataset related entries
+"""
+labels = {
```
tchaton, line 4: All those should dictionnaries be static attributes of the class.

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+        open label file - [semantic label(first 16 bits), instance label(last 16 bits)]
+        returns semantic label, instance label
+        """
+        label = np.fromfile(file_name, dtype=np.uint32)
+        label = label.reshape((-1)) # again for the sake of it
+        semantic_label = label & 0xFFFF
+        instance_label = label >> 16
+        return semantic_label, instance_label
+    
+    def project_scan(self, unproj_points, unproj_remissions):
```
tchaton, line 276: Please, add simple unittests for all transformation functions.

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+        if self.mode == "train":
+            velodyne_file = self.train_scans[index]
+            label_file = self.train_labels[index]
+        elif self.mode == "val":
+            velodyne_file = self.val_scans[index]
+            label_file = self.val_labels[index]
+        else:
+            velodyne_file = self.test_scans[index]
+
+        points, remissions = self.lidar_scan.open_scan(velodyne_file)
```
tchaton, line 403: All this process should happen during a processing step and the scans should be saved as a Data object. Have a look at other dataset's implementation.

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+            v = sorted(glob(os.path.join(velodyne_dir, '*.bin')))
+            velodyne_files.extend(v)
+            if labels:
+                label_dir = os.path.join(self.base_dir, "{0:02d}".format(int(sequence)), "labels")
+                l = sorted(glob(os.path.join(label_dir, '*.label')))
+                label_files.extend(l)
+                if len(v) != len(l):
+                    raise Exception("Sequence {} has unequal label and velodyne files")
+            print("Sequence {} has {} files".format(sequence, len(v)))
+        if labels:
```
tchaton, line 388: Always better to return the same number of arguments. Pad with None.

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+    "img_stds": [12.32, 11,47, 6,91, 0.86, 0.16] 
+}
+
+import numpy as np
+from torch.utils.data.dataset import Dataset
+import os
+from glob import glob
+import torch
+
+class LidarScan:
```
tchaton, line 228: This function looks like a general helper to pointcloud to image. It could be moved to torch_points3d/visualizations/lidar2img.py

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+        proj_x = np.floor(proj_x*self.W) # convert to (0,W) and find nearest integer
+        proj_x = np.minimum(self.W-1, proj_x) # ensure values lie in range
+        proj_x = np.maximum(0, proj_x).astype(np.int32) 
+
+        proj_y = (abs(self.FOV_UP)-pitch)/self.FOV # convert pitch values to 0,1
+        proj_y = np.floor(proj_y*self.H) # convert to (0,H) and find nearest integer
+        proj_y = np.minimum(self.H-1, proj_y) # ensure values lie in range
+        proj_y = np.maximum(0, proj_y).astype(np.int32) 
+
+        # order by decreasing depth
```
tchaton, line 297: Why do you need to order by decreasing depth ?

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+        """
+        self.colormap = np.zeros((max(list(learning_map.keys()))+1, 3), dtype=np.float32)
+        for key, value in color_map.items():
+            value = [value[i] for i in [2,1,0]]
+            self.colormap[key] = np.array(value, np.float32) / 255.0
+        self.reduced_colormap = np.zeros((max(list(learning_map_inv.keys()))+1, 3), dtype=np.float32)
+        for red_cl, map_cl in learning_map_inv.items():
+            self.reduced_colormap[red_cl] = self.colormap[map_cl]
+    
+    def open_scan(self, file_name):
```
tchaton, line 254: could be a @staticmethod

torch_points3d/datasets/segmentation/semantic_kitti.py
```diff
@@ -0,0 +1,441 @@
...
+        open scan file - [x, y, z, remissions]
+        returns point coordinates, remissions
+        """
+        scan = np.fromfile(file_name, dtype=np.float32)
+        scan = scan.reshape((-1, 4)) # just for the sake of it
+        points = scan[:, 0:3]
+        remissions = scan[:, 3]
+        return points, remissions
+    
+    def open_label(self, file_name):
```
tchaton, line 265: could be a @staticmethod

torch_points3d/datasets/segmentation/semantic_kitti.py
tchaton, line 276: could be a @staticmethod and some below too.

torch_points3d/visualization/vispy_vis.py
```diff
@@ -0,0 +1,35 @@
+import vispy
+from vispy.scene import visuals, SceneCanvas
+
+class Vispy:
```
tchaton, line 4: Looks great ! It could take an interface data. Any dataset matching Kitti interface will work with it.
Also, quick question. Where are the data provided ?

tchaton, line 4: Add docstring

torch_points3d/datasets/segmentation/semantic_kitti.py
tchaton, line 360: Add docstring with link to the paper.

torch_points3d/visualization/vispy_vis.py
MukundVarmaT: I did not provide the data here. I left the update function empty as the user can then choose what he/she wants to plot. Like in the code I wrote above I defined my own update function which will set the points and projected images. If lets say the user wants to plot depth images he/she can do that by writing a different update function. I wasn't sure how to generalise so i would love to hear some opinion from you
If its fine to have a visualisation class for semantic kitti then I can provide the data as such

torch_points3d/datasets/segmentation/semantic_kitti.py
MukundVarmaT: I dont think this will come under visualisation because its not just for visualisation but also for building the input output pairs. Since the dataset is huge it may not be viable to actually retain them in memory

humanpose1: I agree with @tchaton

MukundVarmaT:
@humanpose1 @tchaton  I kinda understood what changes are required. So the part inside dataset should just handle the point clouds and the corresponding labels only. All the transformation can be put in data_transforms. The config file should also be shifted to the right place. I will look at the other datasets implemented and try to make it as close as possible to the framework. 

So for the visualisation, I will set up a template for the semantic kitti part. and if let's say someone wants to plot something extra then just create a class calling this base class for vispy and add more stuff. Let me know what you think?

torch_points3d/datasets/segmentation/semantic_kitti.py
nicolas-chaulet: Good idea! Probably reusable across multiple datasets

torch_points3d/visualization/vispy_vis.py
```diff
@@ -0,0 +1,35 @@
+import vispy
```
nicolas-chaulet, line 1: @tchaton @humanpose1 shall we add vispy to the dev dependencies of poetry? `poetry add vispy --dev` or something along those lines...

MukundVarmaT:
Added semantic kitti mode for vispy with the following key press events
- Space to pause continuous running
- After pause
    - up arrow to move forward
    - down arrow to move back
    - space to resume continuous running

Overall would look something like this

vispynew

torch_points3d/visualization/vispy_vis.py
tchaton: Yes, I think so

torch_points3d/datasets/segmentation/kitti.py
```diff
@@ -0,0 +1,221 @@
...
+        label = label.reshape((-1)) # again for the sake of it
+        semantic_label = label & 0xFFFF
+        instance_label = label >> 16
+        return semantic_label, instance_label
+
+    def __getitem__(self, idx):
+        scan = SemanticKitti.open_scan(self._scan_paths[idx])
+        label = None
+        if self._label_path is not None:
+            label, _ = SemanticKitti.open_label(self._label_path[idx])
```
nicolas-chaulet, line 113: You should probably do something here to account for ignored labels in training.

MukundVarmaT: the loss function can have weights which will account for that right?

MukundVarmaT: i will add salsanext which is the current state of the art for segmentation in the semantic kitti dataset. There are point based methods as well and I believe KP conv does pretty well. But right now all of the top methods are dominated by projection based ones.

nicolas-chaulet: Since this labels should be ignored altogether I would recommend to remap the labels just like we do for Scannet. We assign IGNORE_LABEL to the lables we want to ignore and remap the others between 0 and max_labels. Then you can tell the loss which labels should be ignored. The semantic tracker alreadu takes that into account. Well noted for Salsa Next, I was just wondering if there was an easy way to expose a point version of this dataset so that we can have KPConv baselines for example (or Minkowski)

nicolas-chaulet:
Closing for now, feel free to open a new one starting from master that now contains semantic Kitti as a dataset

PR #376, opened 2020-07-30
Descriptors/update 3dm

humanpose1:
Huge PR (sorry ;) ).
- So I updated 3dmatch, it got new functionalities (self supervised mode). I added also add modelnet for registration.
- I've added some transformations for the self supervised methods(like CubeCrop and SphereCrop).
- I've moved the registration utilities (fast global registration and teaser in a torch utilities)
Also I've a scripts for evaluation (it saved the results in a csv file). It's a bit like the evaluate function....

Best


torch_points3d/datasets/registration/modelnet.py
```diff
@@ -0,0 +1,154 @@
...
+from torch_points3d.datasets.classification.modelnet import SampledModelNet
+from torch_points3d.datasets.registration.base_siamese_dataset import BaseSiameseDataset
+from torch_points3d.datasets.registration.base_siamese_dataset import GeneralFragment
+from torch_points3d.metrics.registration_tracker import FragmentRegistrationTracker
+from torch_points3d.datasets.registration.pair import Pair, MultiScalePair
+from torch_points3d.datasets.registration.utils import tracked_matches
+from torch_points3d.datasets.registration.utils import compute_overlap_and_matches
+
+
+class SiameseModelNet(SampledModelNet, GeneralFragment):
```
nicolas-chaulet, line 20: That's super nice

PR #392, opened 2020-08-04
Fixed PointAugment implementation & added PN++ baseline

CCInc:
only tested on PN++


tchaton:
python3 train.py task=classification model_type=point_augment model_name=PointAugmentedModel dataset=modelnet training.batch_size=4

python3 train.py task=classification model_type=pointnet2 model_name=pointnet2_baseline dataset=modelnet training.batch_size=4

tchaton:
```
    result = self.forward(*input, **kwargs)
402
  File "/home/runner/.cache/3.7/lib/python3.7/site-packages/torch/nn/modules/container.py", line 100, in forward
403
    input = module(input)
404
  File "/home/runner/.cache/3.7/lib/python3.7/site-packages/torch/nn/modules/module.py", line 550, in __call__
405
    result = self.forward(*input, **kwargs)
406
  File "/home/runner/.cache/3.7/lib/python3.7/site-packages/torch/nn/modules/container.py", line 100, in forward
407
    input = module(input)
408
  File "/home/runner/.cache/3.7/lib/python3.7/site-packages/torch/nn/modules/module.py", line 550, in __call__
409
    result = self.forward(*input, **kwargs)
410
  File "/home/runner/.cache/3.7/lib/python3.7/site-packages/torch/nn/modules/batchnorm.py", line 106, in forward
411
    exponential_average_factor, self.eps)
412
  File "/home/runner/.cache/3.7/lib/python3.7/site-packages/torch/nn/functional.py", line 1919, in batch_norm
413
    _verify_batch_size(input.size())
414
  File "/home/runner/.cache/3.7/lib/python3.7/site-packages/torch/nn/functional.py", line 1902, in _verify_batch_size
415
    raise ValueError('Expected more than 1 value per channel when training, got input size {}'.format(size))
416
ValueError: Expected more than 1 value per channel when training, got input size torch.Size([1, 512, 1])
```

mingzailao:
I Just Reproduce your code at Brach PointAugment . Nice Work!

Here is the result



but I have some 2 questions to discuss. 

1)
In The Original  paper PointAugment's Appendix : the Matrix $M$ only have 4 parameters not 9


2)
Second Question is : if we Just Set Augmentor Network as GAN:
We need to train a Classifier Well and then adjust the Augmentor? Absolutely follow the algorithm in the original paper, is it Right ?
https://github.com/CCInc/torch-points3d/blob/point_augment/torch_points3d/models/classification/point_augment.py#L61-L69

CCInc:
@mingzailao Thank you for looking into this!

> In The Original paper PointAugment's Appendix : the Matrix $M$ only have 4 parameters not 9

image

I can't find the appendix online, do you have a link to it? In the original paper M is described as a 3x3 matrix which performs scaling and rotation, and a 3D rotation matrix would be a 3x3 shape, so I think it is correct.

> Second Question is : if we Just Set Augmentor Network as GAN:
> We need to train a Classifier Well and then adjust the Augmentor?

I'm not sure what you mean, could you rephrase? For each epoch we optimize the classifier and augmentor separately. It is a bit different from a vanilla GAN because our classification is the main output we are trying trying to optimize.

If you are interested, maybe you can find the issue with the implementation, I'm not sure why it doesn't work very well!

mingzailao:
@CCInc 
Thanks for your reply!



1.[Supplementary material of PointAugment](https://openaccess.thecvf.com/content_CVPR_2020/supplemental/Li_PointAugment_An_Auto-Augmentation_CVPR_2020_supplemental.pdf]
Mention that 


"Note that we produce the shape-wise transformation M by first
output a 4-dimension quaternion and then converting it into a 3 × 3 matrix"
2. As for the second question:
Because we do not have the original code, i just give a little guess. :)

I just find that in training phrase, the classifier always get not good-enough ability, but the augmentor's parameter update quickly, 

for The Algorithm in original paper:




Maybe?
```
Update Augmentor A 1 time

Update Classifier C N time
```

CCInc:
@mingzailao 

We can certainly try outputting a quaternion, though I don't know how much difference it would make, then transform it to a rotation matrix. 

It would take some coding to do more than one forward/backward pass on the classifier. As an alternative, you could try updating the `λ` parameter, in the config file as `aug_weight`.


        # a small λ encourages the augmentor to focus more on the classification with less augmentation on P, and vice versa. 
        # In the paper (all experiments), we set λ = 1 to treat the two terms equally. 

I can't make the changes but if you want to try them I can help guide you! And we can see if they make a difference. Or maybe you can reach out to the author of the paper and see if he can help? Since he hasn't released his code yet.

CCInc:
`nn_rotation: [1024*2, 512, 256, 4, 9]`

I think you are right, this is incorrect. Instead, could you try:

`nn_rotation: [1024*2, 512, 256, 9, 9]`

And it might give a better result?

mingzailao:
@CCInc 
https://github.com/liruihui/PointAugment 
many little tricks 
```python
def batch_quat_to_rotmat(q, out=None):

    B = q.size(0)

    if out is None:
        out = q.new_empty(B, 3, 3)

    # 2 / squared quaternion 2-norm
    len = torch.sum(q.pow(2), 1)
    s = 2/len

    s_ = torch.clamp(len,2.0/3.0,3.0/2.0)

    # coefficients of the Hamilton product of the quaternion with itself
    h = torch.bmm(q.unsqueeze(2), q.unsqueeze(1))

    out[:, 0, 0] = (1 - (h[:, 2, 2] + h[:, 3, 3]).mul(s))#.mul(s_)
    out[:, 0, 1] = (h[:, 1, 2] - h[:, 3, 0]).mul(s)
    out[:, 0, 2] = (h[:, 1, 3] + h[:, 2, 0]).mul(s)

    out[:, 1, 0] = (h[:, 1, 2] + h[:, 3, 0]).mul(s)
    out[:, 1, 1] = (1 - (h[:, 1, 1] + h[:, 3, 3]).mul(s))#.mul(s_)
    out[:, 1, 2] = (h[:, 2, 3] - h[:, 1, 0]).mul(s)

    out[:, 2, 0] = (h[:, 1, 3] - h[:, 2, 0]).mul(s)
    out[:, 2, 1] = (h[:, 2, 3] + h[:, 1, 0]).mul(s)
    out[:, 2, 2] = (1 - (h[:, 1, 1] + h[:, 2, 2]).mul(s))#.mul(s_)

    return out, s_
```

```python
p1 = random.uniform(0, 1)
        possi = 0.5#0.0  
        if p1 > possi:
            pt = torch.bmm(pt, rotation).transpose(1, 2).contiguous()
        else:
            pt = pt.transpose(1, 2).contiguous()
        p2 = random.uniform(0, 1)
        if p2 > possi:
            pt = pt + displacement

```

CCInc:
@mingzailao I will take a look when I have some more time, school is getting very busy now! If you want to try changing some code feel free to add it to this PR. Or if you have any questions we have a slack group for TP3D that you can join, just let me know!

PR #406, opened 2020-08-12
resolve pointnet bugs

tchaton:


tchaton:
I will open a second PR later with a refactored Pointnet API.

PR #408, opened 2020-08-13
Flexible scorer for Pointgroup

nicolas-chaulet:
Some more update on PointGroup with a greater range of scorer that can be switched at train time.


conf/models/panoptic/pointgroup.yaml
```diff
@@ -10,10 +11,143 @@ models:
...
+                stride: 2
+                N: 2
+            up_conv:
+                module_name: ResNetUp
+                dimension: 3
+                up_conv_nn: [[4*in_feat, 2*in_feat], [4*in_feat, in_feat]]
+                kernel_size: 3
+                stride: 2
+                N: 2
+        scorer_encoder:
```
tchaton, line 132: Can you remove scorer_encoder here, as it might bring some confusion. I know it works as architecture: "unet", but it doesn't belong to the point group paper.

conf/training/pointgroup.yaml
```diff
@@ -35,10 +35,10 @@ wandb:
     log: True
     notes: "Minkowski baseline"
     name: "PointGroup"
+    id:
```
tchaton, line 38: ?

torch_points3d/metrics/panoptic_tracker.py
```diff
@@ -277,6 +281,9 @@ def get_metrics(self, verbose=False) -> Dict[str, Any]:
             mAP = sum(self._ap.values()) / len(self._ap)
             metrics["{}_map".format(self._stage)] = mAP
 
+        if verbose:
```
tchaton, line 284: Can you remove the verbose. It is nice to have a per class miou

conf/training/pointgroup.yaml
nicolas-chaulet: This is for resume

torch_points3d/metrics/panoptic_tracker.py
nicolas-chaulet: The verbose just says that it prints that only at the end of an epoch and does not get pushed to wandb (as a graph). It will show in the logs.

conf/models/panoptic/pointgroup.yaml
nicolas-chaulet: Each point group actually has a unet and an encoder scorer at the moment. I did that so I could resume a run with different scorers. I can revert if you want, it was just more convenient.

tchaton: I know why you did it. I meant only for the PointGroupPaper :)

PR #420, opened 2020-08-26
new datasets for registration benchmark

humanpose1:
Hi, I have added new datasets (also, there are slight modifications). 
Also I have added the new metric proposed in https://arxiv.org/pdf/2003.12841.pdf
Code is taken from https://github.com/iralabdisco/point_clouds_registration_benchmark.
I will add the other datasets in an other PR.


nicolas-chaulet (approved):
Looks good! Just some failed tests

humanpose1:
Tests are broken with flake8 and mypy. I don't understand why

PR #427, opened 2020-09-09
correct registration metrics

humanpose1:
After the kind review of @simonefontana see [here](https://github.com/nicolas-chaulet/torch-points3d/pull/420/files#diff-53d25a103d926241da472e184972f24e) for more details, I corrected the metric. It is explained at the section 3.3 of [this article](https://arxiv.org/pdf/2003.12841.pdf).


simone-fontana (approved):
Now it seems correct to me!

nicolas-chaulet (approved):

PR #432, opened 2020-09-17
Add planetary dataset as in https://arxiv.org/abs/2003.12841

simone-fontana:
Similarly to the ETH datasets, I added the planetary dataset, with registration problems defined in https://arxiv.org/abs/2003.12841


torch_points3d/datasets/registration/testplanetary.py
```diff
@@ -0,0 +1,133 @@
...
+import gdown
+import os
+import os.path as osp
+import logging
+import requests
+import glob
+import re
+import sys
+import csv
+import open3d
```
humanpose1, line 14: The test crash, because open3d is not installed if you want to add the package open3d. You have to use poetry.
`poetry add open3d`(If I'm not wrong).  
But whenever you always have to import open3d before torch (or there are some problems)...

humanpose1 (commented):
Overall sounds good. Maybe you should add open3d in poetry and I think it's good

nicolas-chaulet:
When adding dependecies it is better to use `poetry add <package>`.
I'll push a fix for now so that you don't have to it

simone-fontana:
> When adding dependecies it is better to use `poetry add <package>`.
> I'll push a fix for now so that you don't have to it

Yes, sorry for that. In commit 6f82408 I added it manually. However, in the last one I add it with `poetry add open3d@0.10.0.0`

nicolas-chaulet:
Ok, it probably did not update the poetry.lock because the package was already installed

nicolas-chaulet (approved):

PR #435, opened 2020-09-22
Saving dataset properties to the checkpoint

nicolas-chaulet:
This change is required to make sure that a model can be recreated form the checkpoint only. Properties were stored in the dataset but never saved to the checkpoint. They are different from dataset configs in the sense that they capture the properties that model needs from the dataset to be created. Dataset configs are the configs that the dataset need. So I kept them separated.


humanpose1 (approved):
Yes, It's more handy to just have a dictionary for the checkpoints

PR #440, opened 2020-09-24
Pcr benchmark

simone-fontana:
Add Kaist and TUM datasets, as defined in https://arxiv.org/abs/2003.12841, similarly to what we have done for the ETH and Planetary datasets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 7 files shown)


tchaton:
Hey @simone-fontana, could you update readme with new datasets for registration and add your paper to reproduced paper (as all datasets will be added.)  Thanks 👍

nicolas-chaulet:
Yay awesome! Also could you add them in the changelog? All the ones you have added? This will be super useful for our next release

humanpose1 (approved):
Great ! thanks for your contribution

PR #447, opened 2020-10-06
Fix default parameters to elastic distro

nicolas-chaulet:
Updated default parameters to match the ones in MinkowskiEngine


nicolas-chaulet:
@jopetty this should fix the issue you were having right?

humanpose1 (approved):

PR #450, opened 2020-10-06
Registration/kitti dataset

humanpose1:
Finally,
Sorry, I forgot to add kitti odometry.
the download is not automatic yet (I think we are not allowed to download it directly but I'm not sure).
Tell me what you think.


nicolas-chaulet (approved):
Nice!

PR #455, opened 2020-10-08
simple Finetuning

humanpose1:
Just add a small method on base model to easily perform transfert learning
I don't know if you like the interface, tell me what you think !


torch_points3d/models/base_model.py
```diff
@@ -133,6 +134,21 @@ def set_input(self, input, device):
...
+
+        if path_pretrained is not None:
+            if not os.path.exists(path_pretrained):
+                log.warning("The path does not exist, it will not load any model")
+            else:
+                log.info("load pretrained weights from {}".format(path_pretrained))
+                m = torch.load(path_pretrained)["models"][weight_name]
+                self.set_weights(m)
+
+    def set_weights(self, model):
```
nicolas-chaulet, line 149: We don't need that extra method, just use load_state_dict directly

nicolas-chaulet (changes requested):
So if I understand correctly you simply need to specify `path_pretrained` in the model's config and then we load some pretrained weights. What happens when the stored model and the new model have completely different heads for example a pretrained on registration used for sem seg? Does that work?

humanpose1:
Yes, you specify the path in the models config and it will load the pretrained weights, if the file exists. And as it is explained in [this tutorial](https://pytorch.org/tutorials/beginner/saving_loading_models.html#warmstarting-model-using-parameters-from-a-different-model), we can load partial models because the params `strict=False`. But ofc the backbone must have the same name. if the parameters does not have the same name, we have to change manually the names...

nicolas-chaulet:
Sounds good, just the one small change please!

nicolas-chaulet:
(and something in the changelog :))

torch_points3d/trainer.py
```diff
@@ -96,6 +96,7 @@ def _initialize_trainer(self):
         self._checkpoint.dataset_properties = self._dataset.used_properties
 
         log.info(self._model)
+        self._model.set_pretrained_weights()
```
nicolas-chaulet, line 99: Actually, I did not spot that but you need to move that inside the `else` statement just above. Otherwise it will override the weights in `resume` mode (`if not self._checkpoint.is_empty:`)

nicolas-chaulet (changes requested):

torch_points3d/trainer.py
humanpose1: You're right I fix this

nicolas-chaulet (approved):

nicolas-chaulet (approved):

PR #459, opened 2020-10-12
Semantic kitti segmentation

nicolas-chaulet:
Took PR #375 and extracted the dataset for simpler integration. 
It trains fine but it is 3h per epochs with a 10cm grid... 
image


torch_points3d/datasets/segmentation/semantickitti.py
```diff
@@ -0,0 +1,261 @@
...
+            wandb_log - Log using weight and biases
+            tensorboard_log - Log using tensorboard
+        Returns:
+            [BaseTracker] -- tracker
+        """
+        return SegmentationTracker(self, wandb_log=wandb_log, use_tensorboard=tensorboard_log)
+
+
+if __name__ == "__main__":
+    dataroot = "/home/ChauletN/torch-points3d/data/kitti"
```
humanpose1, line 251: that's a good idea !! But It would be better if it's a relative path.
```python
DIR = os.path.dirname(os.path.realpath(__file__))
dataroot = os.path.join(DIR, "..", "..", "..", "..", "data", "kitti")

```

torch_points3d/datasets/segmentation/semantickitti.py
```diff
@@ -0,0 +1,261 @@
...
+            scan_names = []
+            for scan in scan_paths:
+                scan = os.path.splitext(scan)[0]
+                seq, _, scan_id = scan.split(os.path.sep)[-3:]
+                scan_names.append("{}_{}".format(seq, scan_id))
+
+            out_files = [os.path.join(self.processed_paths[i], "{}.pt".format(scan_name)) for scan_name in scan_names]
+            args = zip(scan_paths, label_paths, [self.pre_transform for i in range(len(scan_paths))], out_files)
+            if self.use_multiprocessing:
+                with multiprocessing.Pool(processes=self.process_workers) as pool:
```
humanpose1, line 150: How fast is multi processing ?

humanpose1 (commented):
Nice PR !  Even if KITTI Odometry and Semantic KITTI are from the same dataset, the task is different so the preprocessing is too different.
I'm looking forward to seeing torchpoint3d in the official benchmark !

torch_points3d/datasets/segmentation/semantickitti.py
nicolas-chaulet: Really fast. Basically this part of the process is mainly io, so you have significant speed boosts with that (probably linear with the number of processes)

torch_points3d/datasets/segmentation/semantickitti.py
nicolas-chaulet: Yes good point I used it for debugging

nicolas-chaulet: Probably too many .. no?

humanpose1: It depends on which directory you want to save the data for debug (Maybe I put too much ".." ^^). But it's not so important.

chaitjo:
Quick Questions on SemanticKITTI:
1. Am I correct: The dataset is not automatically downloaded and needs to be extracted to `torch-points3d/torch-points3d/data/kiti/raw/`, i.e. the `rootdir`?
2. As the library doesn't support multi-GPU training yet, how is the performance of trained models compared to SotA? Anecdotally, have you tried re-producing any results reported in papers?

nicolas-chaulet:
Hi! 1. yes you are correct and 2. we haven't :)

chaitjo:
Thanks for the response. Besides KITTI, which segmentation datasets would you say are 'stable'/ready for use with the library if one aims to compare to SotA results? I saw that ScanNet or S3DIS have extensive pre-trained models available in the library; would you recommend using these two over KITTI?

PR #460, opened 2020-10-12
lottery transform

humanpose1:
A transform that chooses randomly a transform among a list of transforms(input is a Omegaconf dict so it use `instantiate_transforms` to build the transforms inside the `LotteryTransform`).


torch_points3d/core/data_transform/__init__.py
```diff
@@ -92,3 +93,35 @@ def instantiate_filters(filter_options):
...
+
+class LotteryTransform(object):
+    """
+    Transforms which draw a transform randomly among several transforms indicated in transform options
+    Examples
+
+    Parameters
+    ----------
+    transform_options Omegaconf list which contains the transform
+
```
nicolas-chaulet, line 107: Remove extra lines

nicolas-chaulet (approved):
haha! Nice one

PR #469, opened 2020-10-22
get rid of warning

fengziyue:
I got this warning when training panoptic segmentation on S3DIS:

`points3d/torch_points3d/datasets/panoptic/utils.py:33: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor)`

So I modified that line.


nicolas-chaulet (approved):
Thanks for the patch!

Issue #476, opened 2020-10-26
Errors on running registration


qshzh:
Thanks for share this amazing work. I want to try torch-points3d on registration tasks. However, I encounter some problems.

1. I encounter the transform errors in the ETHDataset and Kaist. [Log of ETHDataset](https://wandb.ai/qinsheng/default/runs/2mm2pvdo/logs?workspace=user-qinsheng) 
```
  File "/data/torch-points3d/torch_points3d/datasets/registration/basetest.py", line 448, in __getitem__
    return self.get_fragment(idx)
  File "/data/torch-points3d/torch_points3d/datasets/registration/base_siamese_dataset.py", line 142, in get_fragment
    data_source, data_target, new_pair = self.get_raw_pair(idx)
  File "/data/torch-points3d/torch_points3d/datasets/registration/basetest.py", line 444, in get_raw_pair
    data_source_o, data_target_o)
  File "/data/torch-points3d/torch_points3d/datasets/registration/base_siamese_dataset.py", line 122, in unsupervised_preprocess
    data_source = t1(data_source_o.clone())
  File "/data/torch-points3d/torch_points3d/core/data_transform/transforms.py", line 892, in __call__
    data = apply_mask(data, mask)
  File "/data/torch-points3d/torch_points3d/core/data_transform/transforms.py", line 715, in apply_mask
    if(size_pos == len(data[k]) and k not in skip_keys):
TypeError: object of type 'float' has no len()
```

2. Registration performance shown in the [paper](https://arxiv.org/pdf/2010.04642.pdf) looks good. When I run on 3DMatch dataset, there are also some weired error. Such as array not in same shape in the [line](https://github.com/nicolas-chaulet/torch-points3d/blob/e8f99ca60b22ea591e7219cc1989a68e31d955ac/torch_points3d/datasets/registration/pair.py#L218) and RuntimeError: CUDNN_STATUS_MAPPING_ERROR in `MLP2D`. I guess I run with wrong command or maybe miss something important.
It will be great if there are more details on how to reproduce results. 

3. By the way, the auto-download and process in registration datasets have a small bug. The source code of [TORCH_GEOMETRIC Dataset](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/data/dataset.html#Dataset)
```python
        if 'download' in self.__class__.__dict__.keys():
            self._download()

        if 'process' in self.__class__.__dict__.keys():
            self._process()
```
[...]

PR #477, opened 2020-10-26
Registration/fix 476

humanpose1:
Solve some problems in #476.

1) Rearrange the conf file when it is patch based we add _patch in the yaml. 
2) Add in the doc the command to reproduce the results in the paper (still need to complete the tutorial)
3) Solve the missing variable `process` and `download` in `self.__class__.__dict__`. It's a bit hacky but it works :). 

Tell me what you think about it.


torch_points3d/datasets/registration/general3dmatch.py
```diff
@@ -253,6 +254,8 @@ def __init__(
             use_fps=False
     ):
 
+        setattr(self.__class__, "process", self.process)
```
nicolas-chaulet, line 257: that's hacky...
Could you do something like:
```
def process(self):
    super().process()
```

nicolas-chaulet (changes requested):
Minor change requested on style of forwarding the process and download commands

nicolas-chaulet (approved):

PR #489, opened 2020-10-30
Fix problem with batch assembly and grid_size

nicolas-chaulet:


humanpose1 (approved):
Why do we need to store the grid_size ?

nicolas-chaulet:
It’s for the generative network I think. Not sure anymore but we need it
somewhere...

PR #491, opened 2020-11-02
Spvcnn

nicolas-chaulet:


nicolas-chaulet:
@CCInc I have taken your work and integrated it with TP3D now. I have just cherry picked the relevant things for PVCNN since spvnas is not really functional for now. Thanks for initiating this!

torch_points3d/modules/PVCNN/pvcnn.py
```diff
@@ -0,0 +1,164 @@
...
+class PVCNN(nn.Module):
+    def __init__(self, option, model_type, dataset, modules):
+        super(PVCNN, self).__init__()
+
+        cr = option.cr
+        self.vres = option.vres
+        self.num_classes = dataset.num_classes
+        self.num_features = dataset.feature_dimension
+
+        cs = [32, 32, 64, 128, 256, 256, 128, 96, 96]
```
humanpose1, line 24: Maybe we can specify it in the yaml instead of hardcoding it

torch_points3d/modules/PVCNN/utils.py
```diff
@@ -0,0 +1,93 @@
...
+from torchsparse.utils.kernel_region import *
+from torchsparse.utils.helpers import *
+
+
+__all__ = ["initial_voxelize", "point_to_voxel", "voxel_to_point"]
+
+
+# z: PointTensor
+# return: SparseTensor
+def initial_voxelize(z, init_res, after_res):
```
humanpose1, line 15: I think we can do something similar using torch_scatter. Which one is faster ?

torch_points3d/modules/PVCNN/pvcnn.py
```diff
@@ -0,0 +1,164 @@
...
+        x1 = point_to_voxel(x0, z0)
+        x1 = self.stage1(x1)
+        x2 = self.stage2(x1)
+        x3 = self.stage3(x2)
+        x4 = self.stage4(x3)
+        z1 = voxel_to_point(x4, z0)
+        z1.F = z1.F + self.point_transforms[0](z0.F)
+
+        y1 = point_to_voxel(x4, z1)
+        y1.F = self.dropout(y1.F)
```
humanpose1, line 140: does the dropout allow to better generalize here ?

humanpose1 (commented):
Good job !Overall looks good! The architecture is set manually but that's ok if you can reproduce the results of the original paper. If results are good, the next step would be the API !

torch_points3d/modules/PVCNN/pvcnn.py
nicolas-chaulet: Yes, I meant to test it first and then adapt the sparconv3d to allow for this point based branch if it was worth it.

torch_points3d/modules/PVCNN/pvcnn.py
nicolas-chaulet: No idea... Just copy pasted the original network. We will have to run some trainings to figure out what are the best parameters. I may reach out to them because for now it does not look really good

nicolas-chaulet:
Tarining logs:
https://wandb.ai/nicolas/scannet-benchmark/runs/3pc4194r?workspace=user-nicolas

nicolas-chaulet:
I'll just merge it like that. Further exploration would be required as it does not bring better results than a vanilla sparse conv at the moment

chaitjo:
Hi, quick Q: Is the model implemented under pvcnn.py the same as described in the PVCNN paper, or is it the sparse variant SPVCNN? (https://github.com/nicolas-chaulet/torch-points3d/blob/master/torch_points3d/models/segmentation/pvcnn.py)

nicolas-chaulet:
I think the paper actually uses sparse convolutions. So they are one and the same thing from what I understand.

PR #498, opened 2020-11-13
Fix problem with votenet_results.py

daili650:


nicolas-chaulet (approved):
Awesome thanks a lot!!

PR #503, opened 2020-11-22
random Param Transform

humanpose1:
A new transform to select params randomly on a specific transform. (We can use it for a GridSampling with a random grid size)
Example are on the tests. 
What do you think ?

### DIFF
--- a/torch_points3d/core/data_transform/__init__.py
+++ b/torch_points3d/core/data_transform/__init__.py
@@ -131,8 +131,8 @@ class RandomParamTransform(object):
     r"""
     create a transform with random parameters
 
-    Example (on the yaml)::
-
+    Example (on the yaml)
+    '''
     transform: RandomParamTransform
         params:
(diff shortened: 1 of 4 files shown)


torch_points3d/core/data_transform/__init__.py
```diff
@@ -122,3 +125,51 @@ def __repr__(self):
...
+
+
+class RandomParamTransform(object):
+    """
+    create a transform with random parameters
+    Parameters
+    ----------
+
+    transform_options Omegaconf list which contains the transform
+    we need to specify three component for the param (the min, the max and the type)
```
nicolas-chaulet, line 137: Can you an example to the docstring and add that to teh docs?

nicolas-chaulet (changes requested):
Really cool, thanks! Just minor request for the documentation

nicolas-chaulet (approved):

PR #504, opened 2020-11-24
Adding select transform

nicolas-chaulet:


humanpose1 (approved):
thanks for the typo correction !

PR #507, opened 2020-11-28
1. using getattr(data, "norm") instead of data.norm; 2. saving experiment files in a better way

guochengqian:
PyTorch1.7 support. using `getattr(data, "norm", None) is not None` ;
add hydra output config


nicolas-chaulet (approved):
Great! Thanks for the contribution!

PR #511, opened 2020-12-09
Fixing bug when an attribute is missing

nicolas-chaulet:


humanpose1 (changes requested):
There is a problem here. 
When I execute this code:
```python
data = Data()
print(hasattr(data, "norm"))
```
I obtain `True`. because like `pos` and `x` it's set to `None` by default.

nicolas-chaulet:
Ok, that's probably some difference in version of pytorch geometric or pytorch. what about `getattr(data,"norm", None)`?

humanpose1:
```python
print(getattr(data,"norm", None))
```
The result is `None`. So this one is ok. I have the 1.6.1 version of torch geometric.

nicolas-chaulet:
@humanpose1 should be good now.

humanpose1 (approved):
Nice

PR #542, opened 2021-01-24
add pytorch lightning tracker

humanpose1:
I was thinking:
If we want to use pytorch lightning, maybe we need to change the metrics. I propose these modifications.
Tell me what you think about it


nicolas-chaulet (commented):
Going through it quickly, those introduce some breaking changes on our base models correct?

torch_points3d/metrics/classification_tracker.py
```diff
@@ -61,3 +62,22 @@ def metric_func(self):
             "acc": max,
         }  # Those map subsentences to their optimization functions
         return self._metric_func
+
+
+class LightningClassificationTracker(LightningBaseTracker):
```
nicolas-chaulet, line 67: Are these used anywhere? Maybe a way to use those is to reference them from the original trackers and where building the LightningDataModule we instantiate the correct tracker after calling the dataset.get_dataset method?

torch_points3d/utils/colors.py
```diff
@@ -1,7 +1,21 @@
...
 log = logging.getLogger(__name__)
 
+def terminal_size():
+    h, w, _, _ = struct.unpack('HHHH',
+        fcntl.ioctl(0, termios.TIOCGWINSZ,
+        struct.pack('HHHH', 0, 0, 0, 0)))
+    return w, h
+
+
+def breakpoint_zero():
```
tchaton, line 16: Can be removed.

tchaton (approved):
Looks really good !

humanpose1:
@nicolas-chaulet  @tchaton Thanks a lot for your useful feedback :) . Right now I'm a bit busy but I will take into account your remarks to have a cleaner PR and finish the integration of PL.

@nicolas-chaulet , Yes I have modified the `BaseModel` and every models. basically, every `BaseModel` have a loss attribute so the backward is similar for every models (it allows to avoid repetitions).

PR #551, opened 2021-02-10
updatedeps

nicolas-chaulet:
Some overdue update :)


humanpose1 (approved):
Thanks for the update ! looks good to me. Have you tried Minkowski models in `PretrainedRegistry` ?

nicolas-chaulet:
Not beyond what we have in the tests. So should be good!

Le mer. 24 févr. 2021 à 15:14, humanpose1 <notifications@github.com> a
écrit :

PR #559, opened 2021-03-07
Hydra 1.0

omry:
The biggest problem is the usage of the non-feature of combining missing (???) with optional : true.
A better way to achieve the same result is to create config groups for task, dataset and model_type which will contain empty files with named after valid choices.

OmegaConf 2.0 (that comes with Hydra 1.0) has a bug that prevent the usage of empty files, so a dummy _empty_: true field is added.

The rest of the changes are introducing `@package` header to all configs to eliminate the deprecation warnings.
Many of the package headers will be removed with Hydra 1.1, but now now this is the way to go.

At a high level, I think the config structure can use some rethinking but this is way beyond the scope of this diff.

Notes:
1. I compare the results to the config generated by hydra 0.11 and it's the same with the exception of the `_emtpy_: true` fields.
2. This is not a comprehesive diff, I am pretty sure some more choices should be added to task, dataset and model_type (I added just a few to make sure things looks okay).

I suggest that you take this diff as a base for a real diff and do not merge it as is.


nicolas-chaulet:
@omry thank you soo much for this proposal, we will take a look! 

> I am pretty sure some more choices should be added to task, dataset and model_type (I added just a few to make sure things looks okay).

Do you mean that until we add all the datasets and models as "empty" packages they would break?

omry:
> @omry thank you soo much for this proposal, we will take a look!
> 
> > I am pretty sure some more choices should be added to task, dataset and model_type (I added just a few to make sure things looks okay).
> 
> Do you mean that until we add all the datasets and models as "empty" packages they would break?

They would just be counted as valid choices.
You will get a proper error telling you what the accepted choices are, check this out and try.

nicolas-chaulet:
I gave it a try and it's certainly nice that it lists the configs that area available when you try to do something that's not there. At the moment we actually handle that in the code.
The main challenge in my opinion is that it is one more step to not "forget" when creating a new model or dataset. Is there any way we could avoid that? It would be ok since we do the check in the code anyway. 

FYI @humanpose1 every time you create a new dataset or model you would have to add a file to the `conf/dataset` or `conf/model_type` folders.

nicolas-chaulet:
I think it is actually confusing to have a `dataset` folder and a separate `data` folder. Does not feel right to me. Could we somehow load those "allowed" options from a single file? That would help already.

omry:
> The main challenge in my opinion is that it is one more step to not "forget" when creating a new model or dataset. Is there any way we could avoid that? It would be ok since we do the check in the code anyway.

That is a small challenge, Hydra will tell you that you forgot when you try to run it by indicating that the choice in not in the config group.

> I think it is actually confusing to have a dataset folder and a separate data folder. Does not feel right to me. Could we somehow load those "allowed" options from a single file? That would help already.

You can add the equivalent configs to the config store instead.

I agree with you that the structure is weird, as I said you should rethink the config structure.
At a glance, I see some near duplications of configs in your structure (e.g conf/data/object_detection/scannet-sparse.yaml and conf/data/panoptic/scannet-sparse.yaml) which makes me wonder if there isn't a better way to express your config by reusing in this example scannet-sparse.yaml from once place (conf/dataset).

Hydra 1.1 (still in development) will add the ability to have a Defaults List in all config files which will add a ton of new flexibility that will enable more creative ways to solve some of the problems you have.
Even looking with 1.0 (and 0.11) in mind, I think your config structure is somewhat inefficient.

Speaking of 1.1, I strongly recommend that you upgrade to 1.0 first (Upgrade will be MUCH easier if you do not skip major versions).

omry:
The config store is described in the Structured Configs tutorial.
https://hydra.cc/docs/tutorials/structured_config/intro

You can store plain dictionaries into it as well.

nicolas-chaulet:
Yes I know the config system has duplicates. It is by design in a way and all stems from the split between tasks. They should all be self-contained and independent in their config system. I'll take a look at the config store thanks for the pointer.

omry:
This your choice to make, but I think it's wrong :).

One extreme is to have a single config for each combination. 100% duplication and 100% safety.
On the other hand you can strive to minimize duplication and maximize reuse.
Hydra, is providing more and more tools to do that while retaining reproducibility.

In most scenarios you should be able to express your config without duplication. This makes life much easier as things are evolving - reducing the need to update things in many places.

nicolas-chaulet:
Closing in favour of #604 
Thanks @omry for the suggestions!

PR #573, opened 2021-03-24
Update Scheduler

### LINKED ISSUE #487: Scheduler raises a warning
UserWarning: The epoch parameter in `scheduler.step()` was not necessary and is being deprecated where possible. Please use `scheduler.step()` to step the scheduler. During the deprecation, if epoch is different from None, the closed form is used instead of the new chainable form, where available. Please open an issue if you are unable to replicate your use case: https://github.com/pytorch/pytorch/issues/new/choose.
  warnings.warn(EPOCH_DEPRECATION_WARNING, UserWarning)
**why this warnings?**

CCInc:
Trying to fix  #487. The change steps the scheduler however many times. 

Note that the scheduler that updates on samples changes the learning rate before the epoch scheduler. The epoch scheduler has to update after the first step of the new epoch, while the sample scheduler updates after the last step of the current epoch. This was avoided before by setting `_num_samples` to `-1`, but this no longer works. 

Let me know any thoughts/suggestions if there's a better way to implement this!

### DIFF
--- a/test/test_lr_scheduler.py
+++ b/test/test_lr_scheduler.py
@@ -11,7 +11,7 @@ ROOT = os.path.join(DIR, "..")
 sys.path.insert(0, ROOT)
 
 from torch_points3d.models.base_model import BaseModel
-from mock_models import DifferentiableMockModel
+from test.mock_models import DifferentiableMockModel
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


nicolas-chaulet (approved):
Looks great to me, thanks a lot for that! It was a long overdue task 🙏

PR #578, opened 2021-04-05
pretrained models from file

humanpose1:
Simple function to quickly and directly load a model from a local file.


tchaton (approved):
LGTM ! Should we add a test ?

PR #604, opened 2021-05-11
Moving towards Hydra 1.1

liubigli:
Hi everybody,

I am opening this pull request because I would like to have a discussion on how to move to Hydra 1.1
I started from the previous pull request that proposed a first update of config files : https://github.com/nicolas-chaulet/torch-points3d/pull/559 . However, in this pr I don't like the fact that we need to have some redundant empty files such as  for example task/segmentation.yaml

Moreover, @tchaton suggested me to check the implementation in lightning-transformers (https://github.com/PyTorchLightning/lightning-transformers/tree/master/conf). In this case the authors use a config file for each task (e.g. registration, segmentation etc.) and recursive defaults inside several config files. 

Inspired by this last repository I have used the following structure for the conf directory
```bash
├─ conf
    ├─ data            # Dataset conf files
         ├─ classification            
         ├─ object_detection           
         ├─ panoptic            
         ├─ registration            
         ├─ segmentation           
         └─ default.yaml
    ├─ debugging            # Debugging conf files
    ├─ hydra            # Hydra conf files
    ├─ lr_scheduler            # Scheduler conf files
    ├─ models            # Models Conf files
         ├─ object_detection            
         ├─ panoptic            
         ├─ registration            
         ├─ segmentation            
         └─ default.yaml            
    ├─ task            # Task conf files
         ├─ classification.yaml            
         ├─ default.yaml            
         ├─ object_detection.yaml            
         ├─ panoptic.yaml            
         ├─ registration.yaml            
         └─ segmentation.yaml           
    ├─ training            # Training conf files
    ├─ visualization            # Visualisation conf files
    ├─ config.yaml            # General config file
    ├─ eval.yaml            # Eval config file
    └─ sota.yaml            
```

To be honest I have some concerns about the proposed solution, mostly because I had to change the arguments of the call to train function. For example, with the proposed solution we run the train function using the command: 
```bash
poetry run python train.py task=segmentation models=segmentation/pointnet2 model_name=pointnet2_charlesssg data=segmentation/shapenet-fixed
```
 However, I think that it could be nice to start moving to a solution that uses recursive defaults and this is a first step towards.
Thank you in advance

P.S. For tests, I have used hydra==1.1.0dev6 and omegaconf==2.1.0dev26


torch_points3d/datasets/segmentation/scannet.py
```diff
@@ -815,10 +815,11 @@ def __init__(self, dataset_opt):
 
         use_instance_labels: bool = dataset_opt.use_instance_labels
         use_instance_bboxes: bool = dataset_opt.use_instance_bboxes
-        donotcare_class_ids: [] = dataset_opt.donotcare_class_ids if dataset_opt.donotcare_class_ids else []
-        max_num_point: int = dataset_opt.max_num_point if dataset_opt.max_num_point is not None else None
-        process_workers: int = dataset_opt.process_workers if dataset_opt.process_workers else 0
-        is_test: bool = dataset_opt.is_test if dataset_opt.is_test is not None else False
+        # donotcare_class_ids: [] = dataset_opt.donotcare_class_ids if dataset_opt.donotcare_class_ids else []
+        donotcare_class_ids: [] = list(dataset_opt.donotcare_class_ids) if hasattr(dataset_opt,'donotcare_class_ids') else []
```
nicolas-chaulet, line 819: You could use the .get method instead here and give it [] as a second argument.
This is also applicable throughout this file and other files. Could you please use the .get function everywhere where you've had to fix things for OmegaConf 2.0?

torch_points3d/trainer.py
```diff
@@ -318,3 +320,97 @@ def tracker_options(self):
     @property
     def eval_frequency(self):
         return self._cfg.get("eval_frequency", 1)
+
+class Instantiator:
```
nicolas-chaulet, line 324: Do you use these two classes somewhere? I could not find it. If not used, could you please remove them?

conf/training/sparse_fragment_reg_finetune.yaml
```diff
@@ -1,48 +0,0 @@
-# Those arguments defines the training hyper-parameters
```
nicolas-chaulet, line 1: Please add this file back unless it has been moved somewhere else?

nicolas-chaulet (changes requested):
Sorry for the lag in replying, that's amazing thanks a lot for putting the time! @CCInc @humanpose1 @tchaton what do you think?

I agree this way of doing requires less boilerplate than the other suggestion that needed someone to create a bunch of empty files every time they would create a new dataset. 

If I understand correctly, the only change that would occur with this new approach is that we need to specify the task in multiple places in the command line (which is acceptable I think). Adding a new dataset config or new model config does not change correct?

I have left some comments, could you please address them? I also updated the poetry files so that CI passes and other people can have the project setup properly. Please make sure you pull the branch before updating your PR.

torch_points3d/trainer.py
liubigli: No I am not using for the moment. I was wondering if we can use this kind of structure in case we want to use pytorch-lightning. Probably is better to leave this for another pull request. I remove it for now.

conf/training/sparse_fragment_reg_finetune.yaml
liubigli: I don't know why but this file was missing from the confs in the pull request https://github.com/nicolas-chaulet/torch-points3d/pull/559 . I add it in the next commit

torch_points3d/datasets/segmentation/scannet.py
liubigli: I have used .get function each time I got an error raised during tests. To avoid any possible error, probably we should use the get function everywhere, but it seems a huge change. Do you want to do it?

liubigli:
Thank you very much @nicolas-chaulet for your comments. I am really happy to have the chance to contribute! 

Exactly! The idea of the solution is to avoid having boilerplate files each time we want to add a dataset or a new model.
At the same time, if we want to move to hydra 1.1 in the future, it could be interesting to use recursive defaults in some config files as the ones in the training folder.

I have done the changes that you requested. Please let me know if you need further changes :)

torch_points3d/datasets/segmentation/scannet.py
nicolas-chaulet: well, maybe just the lines that you have already changed? Like the one above

torch_points3d/trainer.py
nicolas-chaulet: Yes thanks

CCInc:
I think moving towards the new features of hydra/omegaconf like recursive defaults is much needed, along with an overall restructuring of the config files. My initial thoughts would be to restructure the folder tree so that each dataset can be it's own "package" in omegaconf, and we can refactor out boilerplate code from each. For example, most of our dataset configs share common sets of augmentation transforms, so by refactoring those to a common denominator per dataset we can make quick experiment changes without having to change all the files. 
For example, we have 5 segmentation configs involving s3dis which all share the same set of augmentation transforms. In addition, we can refactor the dataset specific settings to a dataset default and override in a config if we need to. In scannet, for instance, we could refactor all of this out:

```
  class: scannet.ScannetDataset
  dataset_name: "scannet-sparse"
  task: segmentation
  dataroot: data
  version: v2
  use_instance_labels: False
  use_instance_bboxes: False
  donotcare_class_ids: []
  process_workers: 4
  use_category: False

  apply_rotation: True
```

So that if I wanted to change the "version" param I wouldn't have to change it in every scannet file.

As for this PR, it seems to be a good start, but let me know how you feel about a greater restructuring. I will read up more on the latest OmegaConf/hydra changes and see if I can help put together a solution with @liubigli .

nicolas-chaulet:
Thanks @CCInc those are great ideas. 

One thing to realise with this PR (and moving to OmegaConf 2.0 is that older checkpoints (that contain OmegaConf objects) will not load due to incompatibility. This for example breaks the pretrained register.

It's not ideal, but one option would be to work some more on this PR to ensure that no Omegacong object gets pickled (only native dicts) and provide a little script to migrate older checkpoints (open them with older omegaconf, convert and omegaconf object to native dict and save). @humanpose1 what are your thoughts?

humanpose1:
Hi everyone, 

Sorry for my late response. Thanks a lot @liubigli  for your contribution. They are great ! I like this update and it's not a lot of efforts for me to add `segmentation/` after each params.

@CCInc  Do you want to have a separate yaml file for the data augmentation ? In your example, a config file would looks like this ? 
```
class: scannet.ScannetDataset
  dataset_name: "scannet-sparse"
  task: segmentation
  dataroot: data
  version: v2
  use_instance_labels: False
  use_instance_bboxes: False
  donotcare_class_ids: []
  process_workers: 4
  use_category: False
  apply_rotation: True
data_augmentation: segmentation/augment1
```

@liubigli  why did you remove the file `conf/data/registration/eth_base.yaml` and `conf/data/registration/eth2_base.yaml` ?
  
I agree with @nicolas-chaulet, backward compatibility is important especially if we want to load the weights because a lot of models have been trained with the old version of omegaconf. But does it mean that we will have to download every pretrained models, change the omegaconf, and load the models again on wandb ? it seems a bit daunting, but I don't have a better idea....
Maybe we can open a new issue in omegaconf's page ?
 
We'll need also to change ModelCheckpoint, to only save Dict files and not omegaconf....

nicolas-chaulet:
Question about backward compatibility is who is using those models (if anyone at all). But if we want to support them, then yes @humanpose1 you are correct, we would need to download, update and upload.

humanpose1:
Some models are important, but I will not be able to convert every models at once. 
IMHO, for this particular problem, the priority is to change `ModelCheckpoint` to be able to save models without omegaconf dependency.  
And when we load the model, it should convert from dict to omegaconf.
But should it be a separate PR ?
Then, we will convert pretrained models with a scripts.

CCInc:
> @CCInc Do you want to have a separate yaml file for the data augmentation ? In your example, a config file would looks like this ?

I don't think having a separate yaml file is needed, since the data augmentations are task- and dataset-specific. And when I say data augmentations, I'm specifically referring to the `RandomNoise`, `RandomRotate`, `RandomScale`, `RandomSymmetry`, etc.  Although it would also be helpful to refactor out common transforms that happen in `train_transforms` and `test_transform`. Conceptually, something like this:

```
├─ conf
    ├─ data            # Dataset conf files         
        ├─ segmentation           
            ├─ scannet         
                ├─ fixed.yaml
                ├─ sparse.yaml
                └─ default.yaml
            └─ default.yaml
         └─ default.yaml
```

conf/data/default.yaml:
```
# @package data
dataroot: /path/
```
conf/data/segmentation/default.yaml:
```
# @package data
defaults:
  - default
task: segmentation
```

conf/data/segmentation/scannet/default.yaml:
```
# @package data
defaults:
  - segmentation/default
class: scannet.ScannetDataset
dataset_name: "scannet-sparse"
version: v2
use_instance_labels: False
use_instance_bboxes: False
donotcare_class_ids: []
process_workers: 4
use_category: False
apply_rotation: True
mode: mean
first_subsampling: 0.05

pre_transform:
- transform: GridSampling3D
  lparams: [0.02]
 
transforms:
    - transform: AddFeatsByKeys
      params:
        list_add_to_x: [True]
        feat_names: ["rgb"]
        delete_feats: [True]
		
augment_transform:
    - transform: RandomNoise
      params:
        sigma: 0.01
        clip: 0.05
    - transform: Random3AxisRotation
      params:
        apply_rotation: ${data.apply_rotation}
        rot_x: 2
        rot_y: 2
        rot_z: 180
    - transform: RandomScaleAnisotropic
      params:
        scales: [0.9, 1.1]
		
train_transform:
	${data.augment_transform}
	${data.transforms}
val_transform: ${data.transforms}
test_transform: ${data.val_transform}
```
conf/data/segmentation/scannet/sparse.yaml:
```
# @package data
defaults:
  - segmentation/scannet/default

grid_size: 0.05
train_transform:
    - transform: ElasticDistortion
	${data.transforms}
    - transform: GridSampling3D
      params:
        size: ${data.grid_size}
        quantize_coords: True
        mode: ${data.mode}

val_transform: 
    ${data.transforms}
    - transform: GridSampling3D
      params:
        size: ${data.grid_size}
        quantize_coords: True
        mode: ${data.mode}
```

Then, the same datasets in other tasks could reference each other to reduce duplication:

conf/data/panoptic/scannet/default.yaml:
```
# @package data
defaults:
  - segmentation/scannet/default
```

Of course, I don't think that syntax is valid (we can't expand the augment_transform and transforms list to make train_transform) but if there's a way to do something similar I think that would be ideal.

I do agree `ModelCheckpoint` should be independent of config files, so that we can still use trained models without relying on a particular config (for ex. if I want to change the transforms a bit between training and testing for some reason)

nicolas-chaulet:
> Some models are important, but I will not be able to convert every models at once.

We support only a few right? It is also acceptable to just say that if people want to use those models they need to use an older TP3d. 

> IMHO, for this particular problem, the priority is to change `ModelCheckpoint` to be able to save models without omegaconf dependency.

Yes 100% and this should probably be done in a different PR

Unless anyone complains, I'll merge this today and we will move forward with this decision.

liubigli:
I am sorry @humanpose1 I forgot to include them. I have used conf files from the previous PR. 
@nicolas-chaulet please let me check if there are any further missing conf files before the merge.

nicolas-chaulet (approved):

CHANGELOG.md
```diff
@@ -5,17 +5,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unrealeased
+
+### Changed
+
+- Migrated to Hydra 1.1 and omegaconf 2.0 [BREAKING any checkpoint created prior to that, in particular the model zoo]
```
omry, line 12: and OmegaConf 2.1.

requirements.txt
```diff
@@ -25,25 +26,26 @@ filelock==3.0.12
...
+gitpython==3.1.16; python_version >= "3.5" and python_full_version < "3.0.0" or python_full_version >= "3.4.0" and python_version >= "3.5"
 google-auth-oauthlib==0.4.4; python_version >= "3.6" and python_full_version < "3.0.0" or python_full_version >= "3.2.0" and python_version >= "3.6"
 google-auth==1.30.0; python_version >= "3.6" and python_full_version < "3.0.0" or python_full_version >= "3.6.0" and python_version >= "3.6"
 googledrivedownloader==0.4; python_version >= "3.6"
 gql==0.2.0; python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.4.0"
 graphql-core==1.1; python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.4.0"
 grpcio==1.37.1; python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.2.0"
 h5py==3.1.0; python_version >= "3.6"
-hydra-core==0.11.3
+hydra-core==1.0.6
```
omry, line 37: I am confused. This is Hydra 1.0, not 1.1.

CCInc:
@omry What are your thoughts on the restructuring I proposed above? Would something like that be doable with hydra 1.1? Any suggestions?

omry:
In general:
1. See if you can move away from using defaults list interpolation in favor of using recursive defaults.
My understanding is that currently the project is binding different config groups choices via defaults list interpolation.
an alternative is to use containing config with a defaults list.
2. See if you can move away from overriding packages via the package header to overriding them via the defaults list.

Hydra 1.1 is very powerful and will likely be able to handle your proposal.
The best way is to try to prototype it and see. If you have concrete questions while you are prototyping it ping us in the Hydra chat and we will try to help.

CCInc:
@omry If you have a chance to review #614, I'd greatly appreciate it. I'm happy with the structure but definitely would be curious if you think I could do something better. I made the two changes you suggested as well.

omry:
@CCInc, sorry- but I don't have the cycles to review it.
If you have specific questions you can ask them in the Hydra chat.

PR #607, opened 2021-05-19
Modelcheckpoint without omegaconf

humanpose1:
- script to convert old model (need the omegaconf version of 1.4.1)
- change ModelCheckpoint to save dict instead of omegaconf DictConfig. 
- resolve some bugs in the trainer

How can we test it automatically ?


torch_points3d/metrics/model_checkpoint.py
```diff
@@ -199,11 +201,11 @@ def start_epoch(self):
 
     @property
     def run_config(self):
-        return self._checkpoint.run_config
+        return OmegaConf.create(self._checkpoint).run_config
```
nicolas-chaulet, line 204: Should not it be `OmegaConf.create(self._checkpoint.run_config) `?

nicolas-chaulet (approved):
Looks good to me, thanks a lot @humanpose1 ! If the script works for you, I am good with that. No need for further testing.

torch_points3d/metrics/model_checkpoint.py
humanpose1: yes

PR #614, opened 2021-05-28
Hydra 1.1 & Config Restructure

CCInc:
Here's my draft for restructuring the configs for hydra 1.1! Datasets configs are split into unique folders and use a recursive defaults structure to reduce duplication. `aug_transform` and `common_transform` are transform lists that are specific to augmentation, and transforms used in both train, val, and test pipelines, respectively. Note that no code was actually added that handles `aug_transform` and `common_transform` - they're strictly used within the config structure to be interpolated into `train/val/test_transform`. If you need more fine-grained control over the train/val/test transforms you can just override those values as normal (ex. `segmentation/scannet/scannet.yaml`  adds a `FixedPoints` transform only to `train_transform`).

CLI Changes:
- No longer have to specify `model_name` or `task`
- `models` -> `model`
- `lr_scheduler` is deprecated, override with `training.optim.lr_scheduler=`
- You can also override the optimizer/bn_scheduler with `training.optim.optimizer=SGD`
- Optimizer settings are defined in `training.optim` to override 

Breaking changes:

- See https://hydra.cc/docs/next/upgrades/intro for how your custom configs might need to be changed
- Changed AddFeatsByKey transform to allow "add_to_x" as optional and default "add_to_x" and "delete_keys" as true
- Config "models" was renamed to "model"
- Each "model" now has its own config file, i.e. the model parameters are direct children of the "model" key. The name of the model is determined by the config file name.
- Removed custom "on_epoch/on_num_sample/on_num_batch" params from lr_scheduler. If you want different parameters for different update schedules, use overrides or hydra defaults list.

TODO:

- [x] Refactor data configs (except for registration, @humanpose1 can do that :))
- [x] Fix lparams in data configs
- [x] Refactor model configs
- [x] Remove `conf/task`
- [x] Update conf/config.yaml
- [x] ~~Rename data configs (`segmentation/scannet/scannet-sparse` -> `segmentation/scannet/sparse`)~~
- [x] Test all configs (`grid_size` vs `first_subsampling` ?)
- [ ] unconverted model confs: model/registration/ms_svconv, segmentation/minkowski
- [ ] update poetry/pip

Let me know if you have any suggestions!


humanpose1 (commented):
I know that specifying the task is now redundant (task=segmentation data=segmentation/scannet models=segmentation/pointnet). However, in your contribution, is it possible to do something weird like this ? (data=segmentation/scannet models=registration/pointnet). 

Also what do you mean by fix lparams ?

Ok, Your design seems better so I'll change it for registration

CCInc:
> I know that specifying the task is now redundant (task=segmentation data=segmentation/scannet models=segmentation/pointnet). However, in your contribution, is it possible to do something weird like this ? (data=segmentation/scannet models=registration/pointnet).

Yes that is possible which is a downside, I might add the "task" parameter back in to avoid that situation.

> Also what do you mean by fix lparams ?

Omegaconf/hydra doesn't like interpolation inside lists using the `[]` syntax. So instead of this:
```
pre_transform:
  - transform: GridSampling3D
    lparams: [${data.grid_size}]
```
we have to do this:
```
pre_transform:
  - transform: GridSampling3D
    lparams: 
        - ${data.grid_size}
```

nicolas-chaulet:
Hey I think I like that! I don't understand how augment_transform are used, I don't see where it's mapped to the train transforms.

CCInc:
It's used in `conf/data/segmentation/default.yaml` by default:

```
val_transform: "${data.common_transform}"
test_transform: "${data.val_transform}"
train_transform: 
  - "${data.aug_transform}"
  - "${data.common_transform}"
```

So users can still directly override `val/train/test_transform` if they want 😄

CCInc:
NOTES:

- Config files that are direct parents to another config files *must* have their `@package` set if the package isn't clear from the folder structure. For example, `s3dis1x1.yaml` inherits from `/data/segmentation/default`, which has a `@package` directive. `s3disfused-addones.yaml` doesn't need a package because it's a leaf. However, the package it inherits from, `s3disfused.yaml` does need a package directive. This is due to the way hydra resolves packages. Not sure if that can be changed. If in doubt, add a `@package` explicitly.
- Similarly, config files that are parents to another config must use a full path or a relative path when specifying their defaults. For example, `s3disfused-addones.yaml` can set its default as a path relative to the package it's in, in this case `segmentation/s3dis/s3disfused`. However, its parent, `s3disfused.yaml` can't set it's default to `segmentation/default`, because it will resolve to `data/segmentation/s3dis/segmentation/default`. Thus, we need to either set its default to a relative path `../default` or specify the full path: `/data/segmentation/default`. Again, not sure if this behavior can be changed. In general, try to use full paths instead of relative paths. When in doubt, just use the full path.
- When you inherit from a parent which has a dictionary, overriding one key of the dictionary will not override the entire dictionary. In contrast, inheriting from a parent which has a list and changing that list will override the entire list. For example:
```
KPConvPaper.yaml: =================

# @package model
defaults:
  - /model/segmentation/default
mlp_cls:
  nn: [ in_feat, in_feat ]
  dropout: 0
  bn_momentum: bn_momentum

KPConvPaper_innermost.yaml: =================

defaults:
- /model/segmentation/kpconv/KPConvPaper

mlp_cls:
  dropout: 0.5

Result: =================

model:
  mlp_cls:
    nn: [ in_feat, in_feat ]
    dropout: 0.5
    bn_momentum: bn_momentum

```

This seems intuitive, but you can run into issues when you need to remove a key from a dictionary. For example, if my `KPConvPaper_innermost` wanted to remove the `mlp_cls.nn` attribute, I would have to explicitly clear it with `nn: {}`.

- To debug your configurations, you should add the `--cfg job` command line argument when calling `train.py`. This argument will only print out the resulting config and stop execution. Try to use the existing config objects as reference when writing your own.
- When possible, numbers that might need to be overridden by a child config should be refactored into their own variables, especially for dataset configs. For example, rather than doing this:

```
common_transform:
  - transform: AddFeatsByKeys
    params:
      feat_names: [pos_z]
```

we should refactor feat_names, which we can then override in a child config without re-typing `common_transform`.
```
features: [ones]
common_transform:
  - transform: AddFeatsByKeys
    params:
      feat_names: ${data.features}
```

- Config files should be refactored to take away as much duplication as possible. For example, in the scannet configs, where there's a lot of parameters we set, we can refactor out all of those parameters into `data/segmentation/scannet/default.yaml`. Note that `default.yaml` dataset files aren't meant to be used directly, but instead used as a base to be inherited by "real" dataset files. 
- `data.common_transform` and `data.aug_transform` aren't used directly by the code. They're simply a way to cleanly refactor code that's used by both train,val, and test transforms (common_transform) and code that's used by only train (aug_transform). If you need more control over this, you can still directly override `train/val/test_transform` in your dataset files. You can see where  `data.common_transform` and `data.aug_transform` are interpolated into `train/val/test_transform` in `data/default.yaml`.
- Remember that configs can use any other config as their parent. For example, `data/object_detection/scannet.yaml` shares most of it's code with `data/segmentation/scannet/scannet.yaml`, so we can just inherit from `data/segmentation/scannet/scannet.yaml` and change the appropriate object_detection specific parameters. Note here that in `data/object_detection/scannet.yaml`, we use the defaults list ordering to our advantage:

```
defaults:
  - /data/segmentation/scannet/scannet
  - /data/object_detection/default
```
First we inherit from the scannet config, then we'll inherit from `object_detection/default` which will override some values in `segmentation/scannet/scannet`.

torch_points3d/models/base_model.py
```diff
@@ -6,6 +6,7 @@
 from torch.optim.optimizer import Optimizer
 from torch.optim.lr_scheduler import _LRScheduler
 import logging
+import hydra
 from collections import defaultdict
 from torch_points3d.core.schedulers.lr_schedulers import instantiate_scheduler
```
nicolas-chaulet, line 11: Please remove those methods since they are not used anymore

train.py
```diff
@@ -3,7 +3,7 @@
 from omegaconf import OmegaConf
 from torch_points3d.trainer import Trainer
 
-
+OmegaConf.register_new_resolver("get_filename", lambda x: x.split('/')[-1])
```
nicolas-chaulet, line 6: Shall this not also be in `eval.py`?

nicolas-chaulet (commented):
This looks great!

train.py
CCInc: Yep I'll have to add those back..

CCInc:
@nicolas-chaulet this will probably be a breaking change for most users, I'm thinking maybe we add this as part of a major release version? maybe 2.0? and do some other major revisions with it as well (ex pytorch lightning), what do you think?

Also, I'm considering redoing the ModelFactory stuff to make it integrate better with hydra, but that'd probably be easier to do once we convert models to pytorch lightning. So I'll finish up this PR and we can put it on a new branch, merge in lightning, and redo the factory.

train.py
CCInc: Added this to the files that use conf.yaml. It's not needed in eval.yaml

nicolas-chaulet:
> release version? maybe 2.0? and do some other major revisions with it as well (ex pytorch lightning), what do you think?

Yes great idea let's create a new 2.0-dev branch and merge that in there. Like that we will be able to make the cut when we're ready. As opposed to breaking main for everyone. @humanpose1 FYI

conf/config.yaml
```diff
@@ -1,23 +1,19 @@
...
-#  - hydra/output: custom  # add the support for user-defined experiment folder (where to save the experiment files)
 
 job_name: benchmark # prefix name for saving the experiment file.
-model_name: ??? # Name of the specific model to load
 update_lr_scheduler_on: "on_epoch" # ["on_epoch", "on_num_batch", "on_num_sample"]
+update_bn_scheduler_on: "on_epoch" # ["on_epoch", "on_num_batch", "on_num_sample"]
 selection_stage: ""
 pretty_print: False
 eval_frequency: 1
+model_name: "${get_filename: ${hydra:runtime.choices.model}}"
```
CCInc, line 16: hacky and breaks `--cfg job --resolve`

conf/data/default.yaml
```diff
@@ -1,2 +1,9 @@
 # @package data
-task: ${task.name}
\ No newline at end of file
+dataroot: data
+use_category: False
```
CCInc, line 3: add default `use_category` bc some models require it

conf/data/default.yaml
```diff
@@ -1,2 +1,9 @@
...
+
+val_transform: "${data.common_transform}"
```
CCInc, line 5: this is where `common_transform` and `aug_transform` get used

conf/model/object_detection/votenet2/VoteNet2Paper.yaml
```diff
@@ -0,0 +1,39 @@
+defaults:
+  - /model/object_detection/default
+  - /model/object_detection/votenet2/backbone@backbone: pn2-paper # override with model/object_detection/votenet2/backbone@model.backbone=rsconv
```
CCInc, line 3: this is really hacky, is there a better way to do this?

conf/training/default.yaml
```diff
@@ -1,27 +1,33 @@
...
+  # lr scheduler params
+  momentum: 0
+  dampening: 0
+  weight_decay: 0
+
+  # bn scheduler params
+  bn_momentum: 0.1
+  bn_decay: 0.9
+  bn_decay_step: 10
+  bn_clip: 1e-2
```
CCInc, line 29: thoughts on this?

test/test_datasets.py
```diff
@@ -11,16 +11,18 @@
 from torch_points3d.models.panoptic.structures import PanopticLabels
 
 
-class TestScannetPanoptic(unittest.TestCase):
-    def test_dataset(self):
-        data_config = OmegaConf.load(os.path.join(DIR_PATH, "test_config/scannet-panoptic.yaml"))
-        dataset = ScannetDataset(data_config.data)
-        self.assertEqual(len(dataset.train_dataset), 2)
-        self.assertEqual(len(dataset.val_dataset), 2)
+# class TestScannetPanoptic(unittest.TestCase):
```
CCInc, line 14: did this test always used to download the entire scannet dataset?

torch_points3d/datasets/segmentation/s3dis.py
```diff
@@ -663,9 +663,6 @@ def __init__(self, dataset_opt):
             transform=self.test_transform,
         )
 
-        if dataset_opt.class_weight_method:
-            self.add_weights(class_weight_method=dataset_opt.class_weight_method)
```
CCInc, line 667: s3dis didn't seem to support class weighting when I tried it

test/test_datasets.py
nicolas-chaulet: It should not. The test folders should contain a dummy version of this dataset with a couple of samples:
https://github.com/nicolas-chaulet/torch-points3d/tree/master/test/data

and the class constructor should see that and not download the dataset. Maybe the change in the config files broke that?

test/test_models.py
```diff
@@ -122,6 +109,9 @@ def is_known_to_fail(model_name):
                 "randlanet",
                 "PVCNN",
                 "ResUNet32",
+                "VoteNet2Paper",
```
nicolas-chaulet, line 112: Changing the config should not increase that list. Is it because some models have not been ported over to the new system?

torch_points3d/datasets/segmentation/s3dis.py
nicolas-chaulet: Actually it should. Maybe something broke since then but we had all the weight logic done and passed to the loss function.

torch_points3d/models/segmentation/ms_svconv3d.py
```diff
@@ -2,7 +2,6 @@
 import torch
 import torch.nn.functional as F
 import torch.nn as nn
-import torchsparse as TS
```
nicolas-chaulet, line 5: Sure about that?

torch_points3d/models/segmentation/sparseconv3d.py
```diff
@@ -1,7 +1,6 @@
 import logging
 import torch.nn.functional as F
 import torch.nn as nn
-import torchsparse as TS
```
nicolas-chaulet, line 4: Sure about that?

nicolas-chaulet (changes requested):
Great work, what's the change in command line argument in the end? How do you launch a training with minimal command line input?

torch_points3d/models/segmentation/ms_svconv3d.py
CCInc: It wasn't being referenced anywhere, shouldn't it use the `SparseConv3d` interface anyways?

torch_points3d/models/segmentation/sparseconv3d.py
CCInc: same as above

test/test_models.py
CCInc: Votenet2 fails because the unittest initializes it without hydra.main and for some reason it can't resolve the defaults list, I don't know how to fix...

default.yaml files shouldn't be used as models

ms_sconv_base hasn't been converted

test/test_datasets.py
CCInc: Fixed

torch_points3d/datasets/segmentation/s3dis.py
CCInc: Fixed, S3DIS Cylinder was the issue

humanpose1 (commented):
Thanks a lot @CCInc  ! I like it . I haven't tested it yet. When It is merged, I will test it for registration.
IMHO, you should also update the `README.md` to explain how to train models with the new version of hydra (also how to debug hydra conf).
But it is not clear to me whether, there is backward compatibility for older models. I mean When we save models, the conf to build the model is in the `run_config`, and I wonder if we will be able to build the model with the "old" `run_config`.

CCInc:
@humanpose1 

>  IMHO, you should also update the README.md to explain how to train models with the new version of hydra (also how to debug hydra conf).

I will definitely be updating the docs and everything once we get everything finalized!

> But it is not clear to me whether, there is backward compatibility for older models. I mean When we save models, the conf to build the model is in the run_config, and I wonder if we will be able to build the model with the "old" run_config.

I don't think there's any technical reason why the models wouldn't work. I think similar to last time, we'll just need to write a conversion script to the new format.

humanpose1:
Ok if it is on a separate branch as @nicolas-chaulet suggested, I can test it to see if older models works.

CCInc:
@humanpose1  I changed the branch to a new branch. Once we merge it it'll be on `nicolas-chaulet:2.0-dev`. All the major changes are documented in the first post, such as changing the model config structure. Also, the `training.optim` config will have to be converted. 

@nicolas-chaulet I added the new CLI command to the first post, let me know your thoughts and if you have any other suggestions/comments!

PR #618, opened 2021-06-03
SparseConv3D Mixed Precision Training Support

CCInc:
I added support for mixed-precision training in torchsparse (https://github.com/mit-han-lab/torchsparse/pull/69) with the following results:

> The performance of the mixed-precision training is almost the same as that of the full-precision training: 76.43 v.s. 76.78.
> The speedup is fairly limited: the total training time is reduced from 7.3 hours to 6.4 hours (around 10% reduction).
> The memory reduction is very significant: the memory usage is reduced from 48.8G to 28.8G (40% reduction).

With these commits, I added a `training.enable_mixed` flag which enables the usage of mixed precision on supported models, as well as a `_supports_mixed` variable to the models which support mixed precision (sp3d with torchsparse backend). 

To actually activate mixed precision, I wrapped the training in `torch.cuda.amp.autocast` and added a loss scaler `torch.cuda.amp.GradScaler`.

I'm running experiments on s3dis which you can find here: https://wandb.ai/allcowseatfood/s3dis 
You can see the mixed precision training has lower memory usage than regular training and autocasted minkowskiengine. 

image

### DIFF
--- a/conf/training/minkowski_scannet.yaml
+++ b/conf/training/minkowski_scannet.yaml
@@ -6,6 +6,7 @@ batch_size: 12
 shuffle: True
 cuda: 0
 precompute_multi_scale: False # Compute multiscate features on cpu for faster training / inference
+enable_mixed: True # Allow mixed precision on supported models
 optim:
   base_lr: 0.1
   # accumulated_gradient: -1 # Accumulate gradient accumulated_gradient * batch_size
--- a/conf/training/minkowski_scannet.yaml
+++ b/conf/training/minkowski_scannet.yaml
@@ -1,5 +1,5 @@
+# @package _global_
 # Ref: https://github.com/chrischoy/SpatioTemporalSegmentation/blob/master/config.py
-# @package training
 epochs: 500
 num_workers: 4
 batch_size: 12
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Migrated to Hydra 1.0 and OmegaConf 2.1 **BREAKING any checkpoint created prior to that, in particular the model zoo**
 
+### Added
+
+- Mixed precision training support for SparseConv3D models with torchsparse backend (Requires torchsparse >= 1.3.0)
+
 ## 1.3.0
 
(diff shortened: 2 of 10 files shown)


conf/training/minkowski_scannet.yaml
```diff
@@ -1,11 +1,12 @@
+# @package _global_
 # Ref: https://github.com/chrischoy/SpatioTemporalSegmentation/blob/master/config.py
-# @package training
```
nicolas-chaulet, line 2: Is this intentional? I thought that this should be `training`

nicolas-chaulet (commented):
Looks great, thanks @CCInc !! 
Could you please add a note in the README maybe under the sparse conv section?
Also, add an entry to the changelog.

nicolas-chaulet:
@CCInc looks like the linter is not super happy, I'll let you fix that and update the doc before approving

CCInc:
Yeah... not sure what to do about that without reworking that whole thing, I think I'll just add `global` to that var so it passes the linter for now.

conf/training/minkowski_scannet.yaml
CCInc: It's kind of broken, `wandb` needs to be in the global namespace, which you can't do with `@package training`. I'll be fixing all this with #614

nicolas-chaulet:
Can't you just deactivate for that line? Adding a global might not be a bad thing, that's what it is in the end...

humanpose1:
That's great ! thanks a lot for your contribution @CCInc !
My data have a lot of points, I'm looking forward to testing mixed precision.
I have a question,  mixed precision is just for training step or it can also be applied for validation or test step (if we want to be faster or less memory footprint) ?

CCInc:
@humanpose1 It should be supported for training and inference, you can also test it by looking at the dtype of the output of the backbone in your models:

```
        features = self.backbone(self.input).x
        print(features.dtype)
```

It should reduce memory footprint a good amount in inference. But, usually from what I see sparseconv networks are usually fully saturated on the computing, so increasing batch size won't get much speed gain (I haven't tried it on anything newer than a Titan X though so it maybe would!)

nicolas-chaulet (approved):
Great, thanks a lot @CCInc !

PR #622, opened 2021-06-14
Bug ms svconv 

humanpose1:
because of Omegaconf, some models could not load on this colab [notebook](https://colab.research.google.com/github/humanpose1/MS-SVConv/blob/main/notebook/demo_MSSVConv.ipynb).

So, I correct it.

### DIFF
--- a/conf/training/sparse_fragment_reg.yaml
+++ b/conf/training/sparse_fragment_reg.yaml
@@ -1,4 +1,3 @@
-# Those arguments defines the training hyper-parameters
 # @package training
 epochs: 200
 num_workers: 6
--- a/conf/training/sparse_fragment_reg_finetune.yaml
+++ b/conf/training/sparse_fragment_reg_finetune.yaml
@@ -1,4 +1,3 @@
-# Those arguments defines the training hyper-parameters
 # @package training
 epochs: 200
 num_workers: 6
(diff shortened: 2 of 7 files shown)


nicolas-chaulet (approved):

nicolas-chaulet:
@humanpose1 there seems to be some trouble with linting, could you please double check? Otherwise looks great!

humanpose1:
Thanks @nicolas-chaulet ! I don't understand why there is this problem of linting...
```
torch_points3d/datasets/registration/testplanetary.py:9: error: Library stubs not installed for "requests" (or incompatible with Python 3.6)
```
normally when we apply mypy we ignore missing imports...

nicolas-chaulet:
Is it working on your local machine?

Le mar. 15 juin 2021 à 09:18, humanpose1 ***@***.***> a
écrit :

humanpose1:
locally if I do:
```
mypy --ignore-missing-imports torch_points3d
```
mypy raises an error
```
torch_points3d/utils/wandb_utils.py:106: error: "None" has no attribute "dir"
Found 1 error in 1 file (checked 206 source files)
```
but it is not the same error of linting...

humanpose1:
I think it comes from the version of mypy. I'm using 0.812 and we used to use vthat version too. Whereas github now seems to use the version 0.902. Maybe now mypy is stricter.

PR #630, opened 2021-06-22
Fix LR Checkpoint Resuming

CCInc:
Previously, when resuming a training run, the optimizer learning rate would reset to the original learning rate rather than the current learning rate. I suspect this is due to the order we load the states: First we initialize the optimizer then set its state to the correct lr, then we initialize the scheduler and set its state (which resets the optimizer to the scheduler's initial value). To fix, we should wait to load the optimizer's state until after we initialize the schedulers.


nicolas-chaulet (approved):
Thanks, that must have been broken at some point

PR #635, opened 2021-07-03
Update dependencies to support torch 1.9

CCInc:
Torchsparse 1.4 changes the api arguments for Conv3d from `transpose` to `transposed`.

`torch-sparse = "^0.6.10"` and `torch-geometric = "^1.7.1"` are needed for torch 1.9 support.

The latest torch-geometric changes some code from `len(dataset)` to `dataset.len()`, which can break with our datasets, so we need to explicitly define the `len` method.


nicolas-chaulet (approved):

nicolas-chaulet:
There is a problem with unittests, it might be coming from python 3.6, could you please remove that and add 3.8? There is no point in supporting 3.6 anymore.

PR #651, opened 2021-08-02
Updating python to 3.8 in docker images

nicolas-chaulet:


CCInc (approved):
Thanks for the fix! Glad we can get pytorch 1.9 integrated

CCInc:
Once we get conda fully working we can probably simplify our docker setup using conda.

PR #668, opened 2021-09-15
Segmentation_MP visualizer fix

### LINKED ISSUE #658: Empty output viz folder
Hi guys, really appreciate the work on torch-points3d, its been super useful for my research piece and creating my own custom model has been pretty seamless so far.

I'm just having an issue with the Visualizer, as it seems that training some models does not output any .ply files in the viz folder (i.e. viz is completely empty).

I am using the visualization/default.yaml config file for all of the tested models with visualization.activate=True.

For example, training pointnet2 with the following command does not output anything in viz:

`python train.py task=segmentation models=segmentation/pointnet2 model_name=pointnet2 data=segmentation/shapenet-fixed`

While training pointnet2_charlesssg does output viz .ply files.

`python train.py task=segmentation models=segmentation/pointnet2 model_name=pointnet2_charlesssg data=segmentation/shapenet-fixed`

Checking the `python train.py --help` configs confirm that for each tested model, the visualization.activate flags are set to True.

Would be great to get clarification if this is a bug or if I've missed some specific config file option somewhere.

Cheers,

Mitch

mitchellwest:
Fixes #658


CCInc:
Hey @mitchellwest , thanks for the fix! It looks good. I was just curious about the precommit change?

mitchellwest:
Hi @CCInc

Ah sorry mistake on the .precommit change, it was causing a bug in my local environment.

I noticed while adding your fix for the visualizer that the `data.pos` variable's  data for dense models is of shape `[B, N, 3]`, whereas for the MP models its `[BxN, 3]`, not sure if this is intended? Had to make a slight modification to the snippet.

CCInc:
Yep that'll be correct! MP and dense tensors will have different shapes.

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     rev: stable
     hooks:
       - id: black
-        language_version: python3.7
+        language_version: python3.8
```
nicolas-chaulet, line 17: @mitchellwest could you please revert that line and push again? After that all is good! Thanks!

nicolas-chaulet (approved):

.pre-commit-config.yaml
nicolas-chaulet: Actually scrap that!

PR #673, opened 2021-09-21
Pytorch profiler

nicolas-chaulet:


torch_points3d/trainer.py
```diff
@@ -136,27 +140,44 @@ def _initialize_trainer(self):
     def train(self):
         self._is_training = True
 
-        for epoch in range(self._checkpoint.start_epoch, self._cfg.training.epochs):
-            log.info("EPOCH %i / %i", epoch, self._cfg.training.epochs)
+        with (torch.profiler.profile(
+            schedule=torch.profiler.schedule(
```
CCInc, line 144: Can we refactor this into its own function to either return a profile or nullcontext?

torch_points3d/trainer.py
```diff
@@ -136,27 +140,44 @@ def _initialize_trainer(self):
...
-                return 0
 
-            if epoch % self.eval_frequency != 0:
-                continue
+            for epoch in range(self._checkpoint.start_epoch, self._cfg.training.epochs):
+                log.info("EPOCH %i / %i", epoch, self._cfg.training.epochs)
 
-            if self._dataset.has_val_loader:
-                self._test_epoch(epoch, "val")
+                with (torch.profiler.record_function('train_epoch') if self.pytorch_profiler_log else nullcontext()):
```
CCInc, line 162: Maybe refactor this too?

CCInc (changes requested):
Awesome!

torch_points3d/trainer.py
ggrzeczkowicz: Done with #679

torch_points3d/trainer.py
ggrzeczkowicz: Done with #679

ggrzeczkowicz:
I refactor torch.profiler.profile and torch.profiler.record_function to use own functions and I move call to profile inside _train_epoch in #679

nicolas-chaulet:
@ggrzeczkowicz looks like static checks are failing, could you please take a look? 
```
make staticchecks
```

ggrzeczkowicz:
I think you could merge that, unless the codecov is a problem.

PR #677, opened 2021-09-23
Updating dev container

nicolas-chaulet:


CCInc:
Our unittests have been broken for a while, any idea why?

nicolas-chaulet:
Yes it is using some older cache. I was trying to upgrade to pytorch 1.8 on this PR but I am having issues with tests failing... Seems like we need to fix a few things here and there

CCInc (approved):

nicolas-chaulet:
yay!!!! @CCInc are you not using pytorch 1.8?

PR #706, opened 2022-01-11
Fix formatting over entire repo

CCInc:
With the addition of precommit-ci, the entire library needs to be formatted back to a good baseline for future PRs.


nicolas-chaulet (approved):
Looks great thanks! Some tests are failing though, I am assuming it is due to mypy being on a newer version or something, would you be able to address that?

CCInc:
Ok, got that fixed. Any thoughts on "Generate requirements.txt"? Should we leave that in or remove it from the precommit?

nicolas-chaulet:
If it fails because requirements.txt is out of date then it's a good thing, the user should update the requirements.txt based on the poetry lock file (there is a command for that if not using pre-commit)

CCInc:
The issue is that it will always fail, because pre-commit.ci doesn't have poetry on their machine. I tried adding it but haven't had any luck. 
image

nicolas-chaulet:
Let's leave it as is then!

PR #738, opened 2022-03-28
Remove obsolete check in RSConvOriginalMSGDown constructor

### LINKED ISSUE #735: Warning: "The down_conv_nn has a different size as radii. Make sure of have SharedRSConv"
The warning in the title is shown when running the RSConv model, however the training still continues without further issues.
I tracked the source of this message down to this:

https://github.com/torch-points3d/torch-points3d/blob/663442df0ca843de3887ab15e74b8a1ac8a0fcd6/torch_points3d/modules/RSConv/dense.py#L276-L277

The two variables are directly coming from the config, which looks like this:

https://github.com/torch-points3d/torch-points3d/blob/663442df0ca843de3887ab15e74b8a1ac8a0fcd6/torch_points3d/applications/conf/rsconv/encoder_4.yaml#L7-L21

As you can see the first conv layer has different kind of arguments. According to the code there is a separate `xyz_raising` layer constructed using the second sublist. I assume that I can ignore this warning and the check above should be extended to handle the first layer in a different way. Is that correct?

ptoews:
Fixes #735 

I digged a bit deeper into it and realized that this check/warning is most likely obsolete, and probably only there because of copy-pasting `RSConvSharedMSGDown`. Here, the `down_conv_nn` argument (or the first sublist in the case of `first_layer`) is immediately unpacked into three values. Therefore, this list must always have a length of three, independent from the radii given.

Because of this I simply removed the check. It should become clear pretty quickly if the length is invalid, since then an exception would tell something like `too many values too unpack, expected 3...`.

### DIFF
--- a/torch_points3d/modules/RSConv/dense.py
+++ b/torch_points3d/modules/RSConv/dense.py
@@ -273,8 +273,6 @@ class RSConvOriginalMSGDown(BaseDenseConvolutionDown):
         **kwargs
     ):
         assert len(radii) == len(nsample)
-        if len(radii) != len(down_conv_nn):
-            log.warning("The down_conv_nn has a different size as radii. Make sure of have SharedRSConv")
         super(RSConvOriginalMSGDown, self).__init__(
             DenseFPSSampler(num_to_sample=npoint), DenseRadiusNeighbourFinder(radii, nsample), **kwargs
         )


CCInc (approved):
Awesome, thanks a bunch for checking it out!

Issue #741, opened 2022-03-31
SparseConv3d: RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation


ptoews:
I'm using the `SparseConv3d` model (`encoder_4.yaml` config) with the Minkowski backend (latest commit from master). During training, I get the following error:

```
File "[...]/venv/lib/python3.7/site-packages/pytorch_lightning/core/lightning.py", line 1434, in backward
    loss.backward(*args, **kwargs)
  File "[...]/venv/lib/python3.7/site-packages/torch/_tensor.py", line 307, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
  File "[...]/venv/lib/python3.7/site-packages/torch/autograd/__init__.py", line 156, in backward
    allow_unreachable=True, accumulate_grad=True)  # allow_unreachable flag
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.FloatTensor [3920, 256]], which is output 0 of ReluBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```

I've tried the anomaly detection thing, however I can't find any changes in the output. Any ideas?

Thanks in advance.

ptoews:
Here is a MWE:

```
import torch
import torch_geometric.data
import torch_points3d.applications.sparseconv3d
import torch_points3d.core.data_transform
import torch.nn.functional as F


t = torch_points3d.core.data_transform.GridSampling3D(0.02, quantize_coords=True)
train_dataset = [
    t(torch_geometric.data.Data(x=torch.randn((1000, 3)),
                                pos=torch.randn((1000, 3)),
                                y=torch.randn((256,))))
    for _ in range(100)
]
train_loader = torch_geometric.data.DataLoader(train_dataset)
model = torch_points3d.applications.sparseconv3d.SparseConv3d("encoder", 3, 4)
optimizer = torch.optim.Adam(model.parameters(), lr=1e-4)

for epoch in range(2):  # loop over the dataset multiple times
    for i, data in enumerate(train_loader, 0):
        # zero the parameter gradients
        optimizer.zero_grad()

        # forward + backward + optimize
        out = model(data)
        y_hat = out.x.squeeze()
        loss = F.mse_loss(y_hat, data.y)
        loss.backward()
        optimizer.step()
```

[...]

PR #757, opened 2022-04-29
fix for issue #741

jaswanthbjk:
The changes fixes the issue #741

### DIFF
--- a/torch_points3d/modules/MinkowskiEngine/api_modules.py
+++ b/torch_points3d/modules/MinkowskiEngine/api_modules.py
@@ -74,11 +74,11 @@ class ResBlock(ME.MinkowskiNetwork):
             self.downsample = None
 
     def forward(self, x):
-        intermediate_out = self.block(x)
+        out = self.block(x)
         if self.downsample:
-            out = intermediate_out + self.downsample(x)
+            out = out + self.downsample(x)
         else:
-            out = intermediate_out + x
+            out = out + x
         return out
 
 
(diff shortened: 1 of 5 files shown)


torch_points3d/modules/MinkowskiEngine/api_modules.py
```diff
@@ -74,11 +74,11 @@ def __init__(self, input_nc, output_nc, convolution, dimension=3):
             self.downsample = None
 
     def forward(self, x):
-        out = self.block(x)
+        intermediate_out = self.block(x)
```
CCInc, line 77: I think you can leave the variables named `out` throughout all of this

torch_points3d/modules/MinkowskiEngine/api_modules.py
```diff
@@ -201,12 +201,12 @@ def __init__(self, input_nc, output_nc, convolution, dimension=3, reduction=16):
         self.SE = SELayer(output_nc, reduction=reduction, dimension=dimension)
 
     def forward(self, x):
-        out = self.block(x)
-        out = self.SE(out)
+        intermediate_out_1 = self.block(x)
```
CCInc, line 204: Same thing, I think these can remain as `out`

torch_points3d/modules/MinkowskiEngine/res16unet.py
```diff
@@ -46,7 +46,7 @@ def forward(self, x):
         if self.downsample is not None:
             residual = self.downsample(x)
 
-        out += residual
+        out = out.clone() + residual
```
CCInc, line 49: Do you know if the `.clone` is necessary? Will it work if you just do:

`out = out + residual`

CCInc (commented):
Thanks so much for the PR!

I think there may also need to be some files changed in https://github.com/torch-points3d/torch-points3d/blob/master/torch_points3d/modules/MinkowskiEngine/resunet.py similar to in res16unet.py, would you mind adding the fix to that file as well?

jaswanthbjk:
@CCInc I had done the changes specified and also some additional changes to SparseConv module

CCInc (approved):
Thanks so much!!

